PR #12, opened 2017-11-14
Check for Location Permission

electrostat:
Before requesting location updates through the location engine, check to see if the application already has location permission. If not, prevent location updates from running and fall back on an exponential backoff system for checking on the status of the location permission. Once permission checks pass, allow location engine to operate normally.


liblocation/src/main/java/com/mapbox/services/android/core/permissions/ExponentialBackoff.java
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Returns the elapsed time in milliseconds since an {@link ExponentialBackoff} instance is
+   * created and is reset when {@link #reset()} is called.
+   *
+   * <p>
+   * The elapsed time is computed using {@link System#nanoTime()}.
+   * </p>
+   */
+  public final long getElapsedTimeMillis() {
+    return (System.nanoTime() - startTimeNanos) / 1000000;
```
Guardiola31337: NIT Magic number

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
```diff
@@ -184,7 +188,7 @@ private void sendEvents(List<Event> events, Callback httpCallback) {
   }
 
   private boolean startTelemetry() {
-    if (!isTelemetryEnabled) {
+    if (!isTelemetryEnabled && checkLocationPermission()) {
```
Guardiola31337: Adding the check here would mean that https://github.com/mapbox/mapbox-events-android/blob/a2ec6948397250aa784ee9f4e478c64d646c5eb3/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java#L190-L192 won't be executed unless the user has the permissions granted and we don't want that.

The expected behavior is to have everything working except the location-related stuff.
We added the following `// TODO` comments to analyze if it would make sense to add the check within `TelemetryService` or not:
https://github.com/mapbox/mapbox-events-android/blob/a2ec6948397250aa784ee9f4e478c64d646c5eb3/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryService.java#L39
https://github.com/mapbox/mapbox-events-android/blob/a2ec6948397250aa784ee9f4e478c64d646c5eb3/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryService.java#L51
The idea of not creating the service if the permissions are not granted is great but we should keep the rest of functionalities working.

Now it's a good moment to extract that block of code into a well-named method.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryService.java
```diff
@@ -167,4 +175,33 @@ TelemetryService obtainService() {
...
+    if (PermissionsManager.areLocationPermissionsGranted(this)) {
+      return true;
+    } else {
+      permissionBackoff(this);
+      return false;
+    }
+  }
+
+  private void permissionBackoff(final Context context) {
+    final Handler handler = new Handler();
```
Guardiola31337: This code may cause a memory leak.
👀 [How to Leak a Context: Handlers & Inner Classes](http://www.androiddesignpatterns.com/2013/01/inner-class-handler-memory-leak.html)

liblocation/src/main/java/com/mapbox/services/android/core/permissions/ExponentialBackoff.java
```diff
@@ -0,0 +1,118 @@
+package com.mapbox.services.android.core.permissions;
+
+/**
+ * Simplified version of Google HTTP Client Library for Java's ExponentialBackOff.
+ * https://developers.google.com/api-client-library/java/google-http-java-client/backoff
+ *
+ * In this version, the interval creation is deterministic and has no concept of
+ * maxElapsedTimeMillis, it's up to the application to stop. It also adjusts its default values.
+ */
+public class ExponentialBackoff {
```
Guardiola31337: This class is only used internally, so no need to be `public` (we don't need to expose it publicly). The same thing happens with the properties and the methods included.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
```diff
@@ -196,9 +200,11 @@ private boolean startTelemetry() {
 
   private boolean optLocationIn() {
     if (isTelemetryEnabled && !isOpted) {
-      startLocation();
       registerEventReceiver();
       isOpted = true;
+      if (checkLocationPermission()) {
```
Guardiola31337: Adding the check here would mean that
https://github.com/mapbox/mapbox-events-android/blob/22f1465aebbc21bb4924a6cfbf8a3879e1592754/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java#L203-L204 will be executed and we don't want that.

No need to register the receiver in charge of dealing with the location probes if permissions are not granted. Besides that, `isOpted` should only be `true` if previous actions `startLocation()` and `registerEventReceiver()` were executed, meaning that location is opted. If not, we may have issues when `optLocationOut()`.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
```diff
@@ -277,4 +283,43 @@ public void onServiceDisconnected(ComponentName className) {
...
+    Handler handler = new Handler();
+    ExponentialBackoff counter = new ExponentialBackoff();
+
+    PermissionCheckRunnable permissionCheckRunnable = new PermissionCheckRunnable(context);
+
+    long nextWaitTime = counter.nextBackOffMillis();
+    handler.postDelayed(permissionCheckRunnable, nextWaitTime);
+  }
+
+  private final class PermissionCheckRunnable implements Runnable {
```
Guardiola31337: Why is `PermissionCheckRunnable` a nested class of `MapboxTelemetry`? Is it necessary? Could we extract `PermissionCheckRunnable` class into a `PermissionCheckRunnable.java` file/class?

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryService.java
```diff
@@ -48,8 +47,10 @@ public int onStartCommand(Intent intent, int flags, int startId) {
 
   @Override
   public void onDestroy() {
-    // TODO If the location receiver was not created (permissions were not granted) do not unregister
-    unregisterLocationReceiver();
+    if (locationReceiver != null) {
```
Guardiola31337: I think this is not necessary because if we never create `TelemetryService` we can't get a `null` `locationReceiver`. This is because how we're dealing with the `TelemetryService` creation in `MapboxTelemetry`. Am I missing something?

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
```diff
@@ -277,4 +278,29 @@ public void onServiceDisconnected(ComponentName className) {
...
+
+    PermissionCheckRunnable permissionCheckRunnable = new PermissionCheckRunnable(context, this);
+
+    long nextWaitTime = counter.nextBackOffMillis();
+    handler.postDelayed(permissionCheckRunnable, nextWaitTime);
+  }
+
+  void locationAndReceiverSetup() {
+    registerEventReceiver();
+    isOpted = true;
```
Guardiola31337: https://github.com/mapbox/mapbox-events-android/pull/12#discussion_r154335065
> Besides that, `isOpted` should only be `true` if previous actions `startLocation()` and `registerEventReceiver()` were executed, meaning that location is opted.

`startLocation()` and `registerEventReceiver()` might fail, order matters.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/PermissionCheckRunnable.java
```diff
@@ -0,0 +1,28 @@
...
+
+  PermissionCheckRunnable(Context context, MapboxTelemetry mapboxTelemetry) {
+    this.context = context;
+    this.mapboxTelemetry = mapboxTelemetry;
+  }
+
+  @Override
+  public void run() {
+    if (PermissionsManager.areLocationPermissionsGranted(context)) {
+      mapboxTelemetry.locationAndReceiverSetup();
```
Guardiola31337: If `isTelemetryEnabled` or `isOpted` change their value between intervals will cause problems.
No need to create `locationAndReceiverSetup` method because we could use the `public` method `optIn` which will check those variables too.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
```diff
@@ -277,4 +278,29 @@ public void onServiceDisconnected(ComponentName className) {
...
+    if (PermissionsManager.areLocationPermissionsGranted(context)) {
+      return true;
+    } else {
+      permissionBackoff();
+      return false;
+    }
+  }
+
+  private void permissionBackoff() {
+    Handler handler = new Handler();
```
Guardiola31337: There's no need to create a handler and a counter here because we're already doing it in `PermissionCheckRunnable` (it may be interesting to pass them in through the constructor though).

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
```diff
@@ -277,4 +278,29 @@ public void onServiceDisconnected(ComponentName className) {
...
+    ExponentialBackoff counter = new ExponentialBackoff();
+
+    PermissionCheckRunnable permissionCheckRunnable = new PermissionCheckRunnable(context, this);
+
+    long nextWaitTime = counter.nextBackOffMillis();
```
Guardiola31337: No need to calculate `nextWaitTime` and calling `handler.postDelayed(permissionCheckRunnable, nextWaitTime)` because of ☝️
What about calling `permissionCheckRunnable.run()` instead?

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/PermissionCheckRunnable.java
```diff
@@ -0,0 +1,28 @@
...
+
+  PermissionCheckRunnable(Context context, MapboxTelemetry mapboxTelemetry) {
+    this.context = context;
+    this.mapboxTelemetry = mapboxTelemetry;
+  }
+
+  @Override
+  public void run() {
+    if (PermissionsManager.areLocationPermissionsGranted(context)) {
+      mapboxTelemetry.optIn();
```
Guardiola31337: I guess this won't work properly because `mapboxTelemetry.optIn()` 👉 `optLocationIn()` 👉 `checkLocationPermission()` 👉 `permissionBackoff()` which will create a new instance of `PermissionCheckRunnable` so the intervals will be reset (loosing the exponential backoff behavior).
What about creating a lazy initialization method for `PermissionCheckRunnable` creation?

libtelemetry/src/test/java/com/mapbox/services/android/telemetry/MapboxTelemetryTest.java
```diff
@@ -473,4 +473,37 @@ public void checksUnregisterReceiverWhenOptedOut() throws Exception {
 
     verify(mockedLocalBroadcastManager, times(1)).unregisterReceiver(eq(theEventReceiver));
   }
+
+  @Test
+  public void checksLocationPermissionSet() throws Exception {
```
Guardiola31337: This is the same exact test as the `checksOptedIn` one. We should remove it.

libtelemetry/src/test/java/com/mapbox/services/android/telemetry/MapboxTelemetryTest.java
```diff
@@ -473,4 +473,37 @@ public void checksUnregisterReceiverWhenOptedOut() throws Exception {
...
+    LocalBroadcastManager mockedLocalBroadcastManager = mock(LocalBroadcastManager.class);
+    MapboxTelemetry theMapboxTelemetry = new MapboxTelemetry(mockedContext, aValidAccessToken, mockedEventsQueue,
+      mockedTelemetryClient, mockedHttpCallback, mockedSchedulerFlusher, mockedClock, mockedLocalBroadcastManager);
+    theMapboxTelemetry.enable();
+
+    assertTrue(theMapboxTelemetry.optIn());
+  }
+
+  @Test
+  public void checksLocationPermissionNotSet() throws Exception {
```
Guardiola31337: Here we're not testing the location permissions.
It actually returns `false` because telemetry is not enabled and `isTelemetryEnabled` is `false` https://github.com/mapbox/mapbox-events-android/blob/04732b8d5778e77d2a851632d0a722965e34a7ad/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java#L200
We should revisit this test.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/PermissionCheckRunnable.java
```diff
@@ -0,0 +1,29 @@
+package com.mapbox.services.android.telemetry;
+
+import android.content.Context;
+import android.os.Handler;
+
+import com.mapbox.services.android.core.permissions.PermissionsManager;
+
+class PermissionCheckRunnable implements Runnable {
+  private static PermissionCheckRunnable permissionCheckRunnable;
```
Guardiola31337: > Private field 'permissionCheckRunnable' is never used

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
```diff
@@ -277,4 +280,26 @@ public void onServiceDisconnected(ComponentName className) {
...
+      return true;
+    } else {
+      permissionBackoff();
+      return false;
+    }
+  }
+
+  private void permissionBackoff() {
+    PermissionCheckRunnable permissionCheckRunnable = obtainPermissionCheckRunnable();
+    permissionCheckRunnable.run();
```
Guardiola31337: I think that if we call `permissionCheckRunnable.run()` here the intervals will be reset (losing the exponential backoff behavior), not completely sure though we should test it.
Related 👉 https://github.com/mapbox/mapbox-events-android/pull/12/files#r155230708

electrostat: Tested and works.

PR #20, opened 2017-12-04
Run Telemetry Service on Separate Thread

electrostat:
Find best solution for running clearing the telemetry service from the Main Thread and running it on a separate one. This will improve performance of service, encompassing sdks, and apps.


Guardiola31337:
@electrostat Could we close this PR. It is currently doing nothing.

PR #22, opened 2017-12-07
Debug Logging API

electrostat:
Enables extra logging of telemetry request when activated. This is defaulted to off, but can be activated through a public api.

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
@@ -34,7 +34,6 @@ public class MapboxTelemetry implements FullQueueCallback, EventCallback {
   private boolean isTelemetryEnabled = false;
   private boolean isOpted = false;
   private boolean serviceBound = false;
-  private boolean debugLoggingEnabled = false;
 
   public MapboxTelemetry(Context context, String accessToken, String userAgent, Callback httpCallback) {
     this.context = context;
@@ -97,15 +96,9 @@ public class MapboxTelemetry implements FullQueueCallback, EventCallback {
     }
   }
 
-  public boolean isDebugLoggingEnabled() {
-    return debugLoggingEnabled;
-  }
-
-  public void setDebugLoggingEnabled(boolean debugLoggingEnabled) {
-    this.debugLoggingEnabled = debugLoggingEnabled;
-
+  public void updateDebugLoggingEnabled(boolean debugLoggingEnabled) {
     if (telemetryClient != null) {
-      telemetryClient.setDebugLoggingEnabled(debugLoggingEnabled);
+      telemetryClient.updateDebugLoggingEnabled(debugLoggingEnabled);
     }
   }
 
--- a/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
@@ -124,12 +124,13 @@ public class MapboxTelemetry implements FullQueueCallback, EventCallback {
     }
   }
 
+
   public void updateDebugLoggingEnabled(boolean isDebugLoggingEnabled) {
     if (telemetryClient != null) {
       telemetryClient.updateDebugLoggingEnabled(isDebugLoggingEnabled);
     }
   }
-  
+
   public void updateUserAgent(String userAgent) {
     if (isUserAgentValid(userAgent)) {
       telemetryClient.updateUserAgent(TelemetryUtils.createFullUserAgent(userAgent, context));
--- a/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
@@ -143,7 +143,6 @@ public class MapboxTelemetry implements FullQueueCallback, EventCallback {
   private TelemetryClient createTelemetryClient(String accessToken, String userAgent) {
     TelemetryClientSettings telemetryClientSettings = new TelemetryClientSettings.Builder()
       .environment(Environment.STAGING)
-      .debugLoggingEnabled(false)
       .build();
     telemetryClient = new TelemetryClient(accessToken, userAgent, telemetryClientSettings, new Logger());
 
--- a/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
@@ -96,9 +96,9 @@ public class MapboxTelemetry implements FullQueueCallback, EventCallback {
     }
   }
 
-  public void updateDebugLoggingEnabled(boolean debugLoggingEnabled) {
+  public void updateDebugLoggingEnabled(boolean isDebugLoggingEnabled) {
     if (telemetryClient != null) {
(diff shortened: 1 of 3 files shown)


libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryClient.java
```diff
@@ -48,6 +49,14 @@ public void sendEvent(Event event, Callback callback) throws IOException {
     sendBatch(oneEvent, callback);
   }
 
+  void setDebugLoggingEnabled(boolean debugLoggingEnabled) {
```
Guardiola31337: get/set nomenclature is hijacked by the Java language and should be only used when defining POJOs/DTOs.

In any case, this should be included as a setting of `TelemetryClientSettings`, per consistency with the rest of the settings.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryClient.java
```diff
@@ -48,6 +49,14 @@ public void sendEvent(Event event, Callback callback) throws IOException {
...
+    this.debugLoggingEnabled = debugLoggingEnabled;
+  }
+
+  private boolean isDebugLoggingEnabled() {
```
Guardiola31337: If we include `debugLoggingEnabled` as a setting of `TelemetryClientSettings` ☝️ this won't be necessary.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
```diff
@@ -96,6 +97,18 @@ public void updateSessionIdRotationInterval(int hour) {
     }
   }
 
+  public boolean isDebugLoggingEnabled() {
+    return debugLoggingEnabled;
+  }
+
+  public void setDebugLoggingEnabled(boolean debugLoggingEnabled) {
+    this.debugLoggingEnabled = debugLoggingEnabled;
```
Guardiola31337: We're assigning `debugLoggingEnabled` even if `telemetryClient` is `null` and we don't want that.

In any case we should consider adding `debugLoggingEnabled` setting within `TelemetryClientSettings` 👇 and update it here somehow.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
```diff
@@ -96,6 +97,18 @@ public void updateSessionIdRotationInterval(int hour) {
     }
   }
 
+  public boolean isDebugLoggingEnabled() {
```
Guardiola31337: Why are we exposing `isDebugLoggingEnabled()`? If not needed, let's remove it.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
```diff
@@ -96,6 +97,18 @@ public void updateSessionIdRotationInterval(int hour) {
...
+    return debugLoggingEnabled;
+  }
+
+  public void setDebugLoggingEnabled(boolean debugLoggingEnabled) {
```
Guardiola31337: get/set nomenclature is hijacked by the Java language and should be only used when defining POJOs/DTOs.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
electrostat: Per above description and associated tickets, this is a public api so other sdks and apps can utilize the additional logging to help with testing.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryClient.java
```diff
@@ -56,8 +65,8 @@ private void sendBatch(List<Event> batch, Callback callback) {
     HttpUrl url = baseUrl.newBuilder(EVENTS_ENDPOINT)
       .addQueryParameter(ACCESS_TOKEN_QUERY_PARAMETER, accessToken).build();
 
-    // Extra debug in staging mode
-    if (setting.getEnvironment().equals(Environment.STAGING)) {
+    // Extra debug in staging mode or debug mode
```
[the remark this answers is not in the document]
Guardiola31337: This comment (useless now) should be removed.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryClient.java
```diff
@@ -48,6 +48,10 @@ public void sendEvent(Event event, Callback callback) throws IOException {
     sendBatch(oneEvent, callback);
   }
 
+  void updateDebugLoggingEnabled(boolean debugLoggingEnabled) {
+    setting = setting.toBuilder().debugLoggingEnabled(debugLoggingEnabled).build();
```
Guardiola31337: `setting.toBuilder()` will enter in an infinite loop which will cause a `StackOverflowError` exception 💥 at runtime.

https://github.com/mapbox/mapbox-events-android/blob/2d9ac6b517aab5db6205d27d8765fc0c0ca65004/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryClientSettings.java#L77-L79

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryClientSettings.java
```diff
@@ -107,6 +118,11 @@ public Builder x509TrustManager(X509TrustManager x509TrustManager) {
       return this;
     }
 
+    public Builder debugLoggingEnabled(Boolean debugLoggingEnabled) {
```
Guardiola31337: Why are we using the `Boolean` wrapped form here instead of `boolean`?

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryClientSettings.java
```diff
@@ -68,12 +74,17 @@ private boolean isSocketFactoryUnset(SSLSocketFactory sslSocketFactory, X509Trus
     return sslSocketFactory != null && x509TrustManager != null;
   }
 
+  Builder toBuilder() {
```
Guardiola31337: As indicated before this ends up in an infinite loop causing a `StackOverflowError` exception 💥 at runtime.

We need to implement `toBuilder` method like this one https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java#L344 (👀 the generated code by AutoValue) that would allow us to “copy” all the current values of the builder and only change the ones we’re interested in.
:eyes: https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#-initialize-a-builder-to-the-same-property-values-as-an-existing-value-instance for further information.

libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryClient.java
```diff
@@ -56,8 +60,7 @@ private void sendBatch(List<Event> batch, Callback callback) {
     HttpUrl url = baseUrl.newBuilder(EVENTS_ENDPOINT)
       .addQueryParameter(ACCESS_TOKEN_QUERY_PARAMETER, accessToken).build();
 
-    // Extra debug in staging mode
-    if (setting.getEnvironment().equals(Environment.STAGING)) {
+    if (isExtraDebuggingNeeded()) {
       logger.debug(LOG_TAG, String.format("Sending POST to %s with %d event(s) (user agent: %s) with "
```
Guardiola31337: Magic numbers

PR #32, opened 2018-01-02
Add validation to session ID

electrostat:
Add validation to session identifier update method and only accept integers between 1 and 24.


Guardiola31337:
@electrostat per Slack discussion, I went ahead and submitted https://github.com/mapbox/mapbox-events-android/pull/53 which tackles OP differently. Check it out and feel free to close here if the solution suggested is fine with you.

Guardiola31337:
Closing in favor of #53

PR #39, opened 2018-01-04
Ensure that Location Engine priority is set in the correct range 

electrostat:
Restrict value that can be set for location engine priority, utilizing enum paradigm.


Guardiola31337:
@electrostat per Slack conversation, I went ahead and pushed https://github.com/mapbox/mapbox-events-android/pull/39/commits/432d8e086eb077a3cfb5d273af706df48086eba0 which removes unnecessary `val` from `LocationEnginePriority` `enum` and refactors update priority functionality in the different location engines.

In any case, we'd need to rebase from `master` before landing this PR.

> This branch has conflicts that must be resolved

PR #41, opened 2018-01-05
Require UserId for MapLoadEvent

electrostat:
Spec changed to require UserId for MapLoadEvent

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapEventFactory.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapEventFactory.java
@@ -43,6 +43,7 @@ public class MapEventFactory {
   private static final String NOT_A_LOAD_MAP_EVENT_TYPE = "Type must be a load map event.";
   private static final String NOT_A_GESTURE_MAP_EVENT_TYPE = "Type must be a gesture map event.";
   private static final String MAP_STATE_ILLEGAL_NULL = "MapState cannot be null.";
+  private static final String USER_ID_NULL_EMPTY = "UserId cannot be null or empty.";
   private static final Map<Integer, String> NETWORKS = new HashMap<Integer, String>() {
     {
       put(TelephonyManager.NETWORK_TYPE_1xRTT, SINGLE_CARRIER_RTT);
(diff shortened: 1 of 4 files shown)


libtelemetry/src/main/java/com/mapbox/services/android/telemetry/MapEventFactory.java
```diff
@@ -234,6 +235,12 @@ private void checkLoad(Event.Type type) {
     }
   }
 
+  private void checkUserId(String userId) {
+    if (userId == null && userId.isEmpty()) {
```
Guardiola31337: This would cause a `NullPointerException` 💥 if `userId` is `null`.

> Method invocation 'isEmpty' may produce 'java.lang.NullPointerException'

The IDE hints you about this potential issue/crash.

Guardiola31337: `IllegalArgumentException` should be thrown in two situations:
- If `userId` is `null`
- If `userId` is not `null` but it's empty

Why not using [`TelemetryUtils#isEmpty`](https://github.com/mapbox/mapbox-events-android/blob/master/libtelemetry/src/main/java/com/mapbox/services/android/telemetry/TelemetryUtils.java#L44)?

libtelemetry/src/test/java/com/mapbox/services/android/telemetry/MapEventFactoryTest.java
```diff
@@ -29,7 +29,7 @@ public void checksMapLoadEvent() throws Exception {
     initializeMapboxTelemetry();
     MapEventFactory aMapEventFactory = new MapEventFactory();
 
-    Event mapLoadEvent = aMapEventFactory.createMapLoadEvent(Event.Type.MAP_LOAD);
+    Event mapLoadEvent = aMapEventFactory.createMapLoadEvent(Event.Type.MAP_LOAD, "aUserId");
```
Guardiola31337: This is actually a _valid_ user id.
The same thing happens in the tests below 👇
We should add the missing _not valid_ tests.

PR #43, opened 2018-01-08
Fix hardcoded event version in navigation general metadata

Guardiola31337:
- Fixes hardcoded event version in navigation general metadata adding `eventVersion` parameter to `NavigationMetadata` constructor

👀 @electrostat @zugaldia @danesfeder @ericrwolfe


danesfeder:
👌  @Guardiola31337 @electrostat this will be available in the new library correct?

PR #44, opened 2018-01-08
Make userId required for AppUserTurnstile

electrostat:
Make userId required in constructor for AppUserTurnstile event. This allows clients to add in their own unique Mapbox vendor id.

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java
@@ -28,10 +28,10 @@ public class AppUserTurnstile extends Event implements Parcelable {
   @SerializedName("operatingSystem")
   private String operatingSystem = null;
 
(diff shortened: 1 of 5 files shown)


libtelemetry/src/test/java/com/mapbox/services/android/telemetry/AppUserTurnstileTest.java
```diff
@@ -23,6 +23,8 @@ public void checksAppUserTurnstileType() throws Exception {
 
   private Event obtainAnAppUserTurnstileEvent() {
     boolean indifferentTelemetryEnabled = false;
-    return new AppUserTurnstile(indifferentTelemetryEnabled, "anySdkIdentifier", "anySdkVersion");
+    String userId = TelemetryUtils.obtainUniversalUniqueIdentifier();
```
Guardiola31337: This is not necessary. We could just pass in `"anyUserId"` 👇 and avoid universal unique identifier creation/generation.

The same thing happens with below changed tests.

PR #47, opened 2018-01-10
Store MapboxVendorId

electrostat:
Create a solution for storing and retrieving 'vendorId' string that is used within appUserTurnstile and MapLoad Events

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java
@@ -10,6 +10,8 @@ import com.google.gson.annotations.SerializedName;
 public class AppUserTurnstile extends Event implements Parcelable {
   private static final String APP_USER_TURNSTILE = "appUserTurnstile";
   private static final String OPERATING_SYSTEM = "Android - " + Build.VERSION.RELEASE;
+  private static final String APPLICATION_CONTEXT_CANT_BE_NULL = "Create a MapboxTelemetry instance before calling "
+    + "this method.";
 
   @SerializedName("event")
   private final String event;
@@ -28,10 +30,11 @@ public class AppUserTurnstile extends Event implements Parcelable {
   @SerializedName("operatingSystem")
   private String operatingSystem = null;
 
-  public AppUserTurnstile(boolean enabledTelemetry, String sdkIdentifier, String sdkVersion, String userId) {
+  public AppUserTurnstile(boolean enabledTelemetry, String sdkIdentifier, String sdkVersion) {
+    checkApplicationContext();
     this.event = APP_USER_TURNSTILE;
     this.created = TelemetryUtils.obtainCurrentDate();
-    this.userId = userId;
+    this.userId = TelemetryUtils.retrieveVendorId();
     this.enabledTelemetry = enabledTelemetry;
     this.sdkIdentifier = sdkIdentifier;
     this.sdkVersion = sdkVersion;
@@ -84,4 +87,10 @@ public class AppUserTurnstile extends Event implements Parcelable {
       return new AppUserTurnstile[size];
     }
   };
+
+  private void checkApplicationContext() {
+    if (MapboxTelemetry.applicationContext == null) {
+      throw new IllegalStateException(APPLICATION_CONTEXT_CANT_BE_NULL);
+    }
+  }
 }
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
@@ -1,6 +1,7 @@
 package com.mapbox.android.telemetry;
 
 import android.content.Context;
+import android.content.SharedPreferences;
 import android.content.pm.PackageInfo;
 import android.text.TextUtils;
 
@@ -18,6 +19,8 @@ class TelemetryUtils {
   private static final String THREE_STRING_FORMAT = "%s/%s/%s";
   private static final SimpleDateFormat dateFormat = new SimpleDateFormat(DATE_AND_TIME_PATTERN, Locale.US);
   private static final Locale DEFAULT_LOCALE = Locale.US;
+  private static final String MAPBOX_SHARED_PREFERENCES = "MapboxSharedPreferences";
+  private static final String MAPBOX_SHARED_PREFERENCE_KEY_VENDOR_ID = "mapboxVendorId";
 
   static String obtainCurrentDate() {
     return dateFormat.format(new Date());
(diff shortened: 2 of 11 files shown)


libtelemetry/src/test/java/com/mapbox/android/telemetry/TelemetryClientAppUserTurnstileEventTest.java
```diff
@@ -9,16 +14,19 @@
...
 
+import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class TelemetryClientAppUserTurnstileEventTest extends MockWebServerTest {
 
   @Test
   public void sendsTheCorrectBodyPostingAppUserTurnstileEvent() throws Exception {
+    initializeMapboxTelemetry();
```
Guardiola31337: Why do we need to initialize `MapboxTelemetry`?
We're using `MapboxTelemetry.applicationContext` when retrieving the vendor id from `SharedPreferences` (`TelemetryUtils`) so setting a mocked context should be enough 👀

```java
Context mockedContext = mock(Context.class, RETURNS_DEEP_STUBS);
MapboxTelemetry.applicationContext = mockedContext;
```

libtelemetry/src/test/java/com/mapbox/android/telemetry/TelemetryClientAppUserTurnstileEventTest.java
```diff
@@ -28,4 +36,19 @@ public void sendsTheCorrectBodyPostingAppUserTurnstileEvent() throws Exception {
     String expectedRequestBody = obtainExpectedRequestBody(new GsonBuilder(), theAppUserTurnstile.get(0));
     assertRequestBodyEquals(expectedRequestBody);
   }
+
+  private void initializeMapboxTelemetry() {
```
Guardiola31337: This is not needed per above comment.

libtelemetry/src/test/java/com/mapbox/android/telemetry/TelemetryClientTest.java
```diff
@@ -22,9 +28,15 @@
...
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class TelemetryClientTest extends MockWebServerTest {
+  @Before
+  public void setup() {
+    initializeMapboxTelemetry();
```
Guardiola31337: Similar situation as commented before.
In any case, we should avoid the inline setup and make clear what we're doing/executing on each test.
So, What about including

```java
Context mockedContext = mock(Context.class, RETURNS_DEEP_STUBS);
MapboxTelemetry.applicationContext = mockedContext;
```

or `initializeMapboxTelemetryApplicationContext()` on each test?

`MapboxTelemetry.applicationContext` is a `static` variable and as such it's good to remark that needs to be initialized on each test and setup accordingly to the subject under test making sure that its value is not shared across any other tests (**I**solated from [FIRST principles](http://agileinaflash.blogspot.com.es/2009/02/first.html)).

libtelemetry/src/test/java/com/mapbox/android/telemetry/TelemetryClientTest.java
```diff
@@ -205,4 +217,19 @@ private void assertTelemetryResponseEquals(String responseBody, String expectedM
 
     assertEquals(expectedTelemetryResponse, actualTelemetryResponse);
   }
+
+  private void initializeMapboxTelemetry() {
```
Guardiola31337: This is not needed per above comments.

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
```diff
@@ -61,4 +64,29 @@ private static String obtainApplicationIdentifier(Context context) {
...
     }
   }
+
+  static SharedPreferences getSharedPreferences(Context context) {
+    return context.getSharedPreferences(MAPBOX_SHARED_PREFERENCES, Context.MODE_PRIVATE);
+  }
+
+  static String updateVendorId() {
+    SharedPreferences sharedPreferences = getSharedPreferences(MapboxTelemetry.applicationContext);
+    SharedPreferences.Editor editor = sharedPreferences.edit();
```
Guardiola31337: This is currently doing nothing.
After calling `edit()` on a `SharedPreference`, we must call `commit()` or `apply()` on the editor to save the results. Preferably [`apply()`](https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#apply()).

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
```diff
@@ -61,4 +64,30 @@ private static String obtainApplicationIdentifier(Context context) {
       return EMPTY_STRING;
     }
   }
+
+  private static SharedPreferences obtainSharedPreferences() {
+    return MapboxTelemetry.applicationContext.getSharedPreferences(MAPBOX_SHARED_PREFERENCES, Context.MODE_PRIVATE);
+  }
+
+  static String updateVendorId() {
```
Guardiola31337: This method is only used within `TelemetryUtils` so it should be `private`.

libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java
```diff
@@ -28,10 +28,10 @@
   @SerializedName("operatingSystem")
   private String operatingSystem = null;
 
-  public AppUserTurnstile(boolean enabledTelemetry, String sdkIdentifier, String sdkVersion, String userId) {
+  public AppUserTurnstile(boolean enabledTelemetry, String sdkIdentifier, String sdkVersion) {
```
Guardiola31337: If `MapboxTelemetry.applicationContext` is `null` when creating an `AppUserTurnstile` event we should throw an `IllegalStateException` to avoid potential `NullPointerException`s when calling `TelemetryUtils.retrieveVendorId()`.

Same idea as in `MapEventFactory` 👀 
https://github.com/mapbox/mapbox-events-android/blob/422c54d2d89ea4bda3686640d3204bc6f9696cb0/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapEventFactory.java#L91-L93

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
```diff
@@ -61,4 +64,30 @@ private static String obtainApplicationIdentifier(Context context) {
       return EMPTY_STRING;
     }
   }
+
+  private static SharedPreferences obtainSharedPreferences() {
```
Guardiola31337: NIT We normally organize the code in this particular order :point_right: `public`, `protected`, `private`, etc. trying to locate them in order of calling. Could we restructure the methods of `TelemetryUtils` this fashion?

libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java
```diff
@@ -28,10 +30,13 @@
...
+    if (MapboxTelemetry.applicationContext == null) {
```
Guardiola31337: Could we extract this check into a `private` `checkApplicationContext` method?

PR #56, opened 2018-01-23
Change Packages

electrostat:
Move 4 files into correct package. Issue from rebase.


libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryListener.java
```diff
@@ -1,6 +1,6 @@
-package com.mapbox.services.android.telemetry;
+package com.mapbox.android.telemetry;
 
-public interface TelemetryListener {
+interface TelemetryListener {
```
Guardiola31337: This needs to be `public` so that the clients could add their listeners.

PR #58, opened 2018-01-23
Remove LOST from LocationEngineProvider

electrostat:
Remove lost from LocationEngineProvider to help crash associated with including Lost dependency


zugaldia:
I believe that the reason for this PR is https://github.com/lostzen/lost/issues/250 - do we have any tickets tracking reincluding LOST if that upstream issues gets fixed?

cc: @danesfeder

PR #67, opened 2018-02-20
Refactor Repo

electrostat:
Refactor Main repo to add subdirectories to clean up and condense info as much as possible.

### DIFF
--- a/libtelemetry/gradle.properties
+++ b/libtelemetry/gradle.properties
@@ -1,4 +1,4 @@
-VERSION_NAME=3.1.0-SNAPSHOT
+VERSION_NAME=4.0.0-SNAPSHOT
 POM_ARTIFACT_ID=mapbox-android-telemetry
 POM_NAME=Mapbox Android Telemetry Library
 POM_DESCRIPTION=Mapbox Android Telemetry Library
--- a/libtelemetry/gradle.properties
+++ b/libtelemetry/gradle.properties
@@ -1,4 +1,4 @@
-VERSION_NAME=4.0.0-SNAPSHOT
+VERSION_NAME=3.1.0-SNAPSHOT
 POM_ARTIFACT_ID=mapbox-android-telemetry
 POM_NAME=Mapbox Android Telemetry Library
 POM_DESCRIPTION=Mapbox Android Telemetry Library
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java
@@ -59,7 +59,7 @@ public class AppUserTurnstile extends Event implements Parcelable {
   }
 
   @Override
-  public Type obtainType() {
+  Type obtainType() {
     return Type.TURNSTILE;
   }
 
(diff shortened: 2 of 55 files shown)


electrostat:
@Guardiola31337  These were attempted locally. With Maps Events, `MapDragendEvent` and `MapEventFactory` both require access to TelemetryUtils, which is `Package-Private`. Similar issues arise with Navigation Events and `TelemetryUtils`, also `TelemetryClient.configureGsonBuilder()` would not have access to these events. 

To fix these, we would either need to duplicate functionality in some places and/or open up some classes.

Guardiola31337:
@electrostat 
> These were attempted locally. With Maps Events, `MapDragendEvent` and `MapEventFactory` both require access to TelemetryUtils, which is `Package-Private`. Similar issues arise with Navigation Events and `TelemetryUtils`

We could change `TelemetryUtils` visibility to `public`. Ultimately, it only holds a bunch of `static` methods (utilities) which could be perfectly exposed to the end users. If so, we would need to take care of the `applicationContext` used in some of the methods. A potential solution to workaround this could be to request a `Context` instead of relying on the telem global one.

> also `TelemetryClient.configureGsonBuilder()` would not have access to these events. 

I'm not sure if I follow. `configureGsonBuilder()` is a `private` method inside the package private `TelemetryClient` class, can't 👀 what could be the issue here. Could you clarify?

libtelemetry/gradle.properties
```diff
@@ -1,4 +1,4 @@
-VERSION_NAME=4.0.0-SNAPSHOT
+VERSION_NAME=3.1.0-SNAPSHOT
```
Guardiola31337: This could be perfectly `4.0.0-SNAPSHOT`

libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java
```diff
@@ -59,7 +59,7 @@ public AppUserTurnstile(String sdkIdentifier, String sdkVersion) {
   }
 
   @Override
-  Type obtainType() {
+  public Type obtainType() {
```
Guardiola31337: This needs to be reverted as well.

libtelemetry/src/main/java/com/mapbox/android/telemetry/navigationutils/AudioTypeChain.java
```diff
@@ -1,4 +1,4 @@
-package com.mapbox.android.telemetry;
+package com.mapbox.android.telemetry.navigationutils;
```
Guardiola31337: Could we use the `navigation.utils` package structure as suggested in https://github.com/mapbox/mapbox-events-android/pull/67#pullrequestreview-98119629?

libtelemetry/gradle.properties
electrostat: OK I'm unclear on this. What api breaking change does this do? It is a feature tweak at best. I'm assuming we are going to release many releases before 4.0.0. So I need explanation on your thought process here, because to me, that's jumping the gun.

Guardiola31337: Changing namespaces is a major version because developers need to change their code (pointing to the new package).

Besides that, we normally bump the snapshot version name after a release has happened. It's a way of telling devs what's cooking for the next major version 👀 https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/gradle.properties#L2

We could definitely keep releasing 3.x versions in the meantime (note that this is only for `SNAPSHOT`s).

Does that make sense?

electrostat: 👍

PR #72, opened 2018-03-02
Update OkHttp version to 3.10.0

### LINKED ISSUE #69: Update to OkHttp 3.10.0
OkHttp 3.10.0 has been released and noting from https://github.com/mapbox/mapbox-gl-native/issues/11325 that this introduces a breaking change due to usage of an internal api.

cc @electrostat @Guardiola31337

Guardiola31337:
- Fixes #69 
- Bumps OkHttp version to `3.10.0`
- Adds `hostnameVerifier` to `TelemetryClientSettings` builder needed to workaround stuck MockWebServer tests after bumping OkHttp version to `3.10.0`. For further details 👉 https://github.com/square/okhttp/issues/3898

cc @electrostat @zugaldia @tobrun @LukasPaczos

### DIFF
--- a/libtelemetry/src/test/java/com/mapbox/android/telemetry/MockWebServerTest.java
+++ b/libtelemetry/src/test/java/com/mapbox/android/telemetry/MockWebServerTest.java
@@ -14,6 +14,9 @@ import java.net.HttpURLConnection;
 import java.util.Arrays;
 import java.util.List;
 
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLSession;
+
 import okhttp3.HttpUrl;
 import okhttp3.mockwebserver.internal.tls.SslClient;
 import okhttp3.mockwebserver.MockResponse;
@@ -172,6 +175,12 @@ class MockWebServerTest {
       .baseUrl(localUrl)
       .sslSocketFactory(sslClient.socketFactory)
       .x509TrustManager(sslClient.trustManager)
+      .hostnameVerifier(new HostnameVerifier() {
+        @Override
+        public boolean verify(String hostname, SSLSession session) {
+          return true;
+        }
+      })
       .build();
   }
 }
--- a/libtelemetry/src/test/java/com/mapbox/android/telemetry/MockWebServerTest.java
+++ b/libtelemetry/src/test/java/com/mapbox/android/telemetry/MockWebServerTest.java
@@ -15,7 +15,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import okhttp3.HttpUrl;
-import okhttp3.internal.tls.SslClient;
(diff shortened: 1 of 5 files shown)


libtelemetry/src/test/java/com/mapbox/android/telemetry/MockWebServerTest.java
```diff
@@ -14,8 +14,11 @@
 import java.util.Arrays;
 import java.util.List;
 
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLSession;
+
 import okhttp3.HttpUrl;
-import okhttp3.internal.tls.SslClient;
+import okhttp3.mockwebserver.internal.tls.SslClient;
```
electrostat: Should we continue to use an internal api if future changes could cause issues similar issues? Is there any way we can get rid of this?

libtelemetry/src/test/java/com/mapbox/android/telemetry/TelemetryClientTest.java
```diff
@@ -14,12 +14,15 @@
...
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLSession;
+
 import okhttp3.Call;
 import okhttp3.Callback;
 import okhttp3.HttpUrl;
 import okhttp3.OkHttpClient;
 import okhttp3.Response;
-import okhttp3.internal.tls.SslClient;
+import okhttp3.mockwebserver.internal.tls.SslClient;
```
electrostat: Same as ☝️

libtelemetry/src/test/java/com/mapbox/android/telemetry/MockWebServerTest.java
Guardiola31337: Here is a bit different because MockWebServer is only used for testing. So although APIs could change it's not a problem because it's not affecting production code which is what we're exposing and what could cause issues in the future. Does that make sense?

Guardiola31337: > Is there any way we can get rid of this?

Nope, it's actually needed for testing HTTP requests.

libtelemetry/src/test/java/com/mapbox/android/telemetry/TelemetryClientTest.java
Guardiola31337: Already answered ☝️

libtelemetry/src/test/java/com/mapbox/android/telemetry/MockWebServerTest.java
electrostat: Makes sense. Sounds good.

PR #75, opened 2018-03-02
Location Job Service

electrostat:
Create a job scheduler service to collect background location data


app/src/main/AndroidManifest.xml
```diff
@@ -21,6 +21,14 @@
...
             </intent-filter>
         </activity>
+
+
+        <meta-data
+            android:name="com.mapbox.TestEventsServer"
+            android:value="api-events-staging.tilestream.net"/>
+        <meta-data
+            android:name="com.mapbox.TestEventsAccessToken"
+            android:value="pk.eyJ1IjoiZWxlY3Ryb3N0YXQtdGVzdCIsImEiOiJjamRhaHBhejkydXhlMnhvNmZhZTk3cjI1In0.9I3NYZF29F-XQHW1JIzIPg"/>
```
tobrun: remove hard coded token and put in CI as a secret env. variable

libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationJobService.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import okhttp3.Call;
+import okhttp3.Callback;
+import okhttp3.Response;
+
+@RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+public class LocationJobService extends JobService implements LocationListener, Callback {
```
tobrun: this requires min sdk version of 21, how are we handling this so we don't drop support for lower versions?

PR #77, opened 2018-03-07
Disable Tag

electrostat:
Detect and trigger logic to disable Telemetry events for enterprise developers.


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -486,4 +493,20 @@ private void unregisterEventReceiver() {
     EventReceiver eventReceiver = obtainEventReceiver();
     localBroadcastManager.unregisterReceiver(eventReceiver);
   }
+
+  private boolean isTelemetryDisabled() {
```
Guardiola31337: Could we extract this method and its related variables out of `MapboxTelemetry` into a package-private class? It seems that this behavior could perfectly live inside its own class.

Also, `isTelemetryDisabled` is kind of confusing to me because we have similar method names (`enable`, `disable`, etc.). It'd be great if we could come up with a different (more understandable) name.

PR #100, opened 2018-04-03
Location Engine Singleton Bug

electrostat:
`obtainBestLocationEngine` returns the same instance of a location engine when called. This is causing issues when multiple location engines are used within an app (and created with this method) and getting out of sync with their desired behavior.

cc: @lilykaiser @zugaldia @Guardiola31337


liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProvider.java
```diff
@@ -44,6 +45,7 @@ private void initAvailableLocationEngines(Context context) {
         locationEngineDictionary.put(entry.getKey(), available);
       }
     }
+    Log.e("dropped-location", "locationEngineDictionary: " + locationEngineDictionary);
```
zugaldia: We should probably remove these logs before merging.

app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
```diff
@@ -20,7 +20,7 @@ protected void onCreate(Bundle savedInstanceState) {
     setContentView(R.layout.activity_main);
 
     String accessTokenTelemetry = obtainAccessToken();
-    String userAgentTelemetry = "MapboxEventsAndroid/3.1.0";
+    String userAgentTelemetry = "MapboxEventsAndroid/";
```
zugaldia: This change seems unrelated to this issue?

liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProvider.java
```diff
@@ -57,6 +57,7 @@ private void initAvailableLocationEngines(Context context) {
 
   private LocationEngine obtainBestLocationEngine() {
     LocationEngine androidLocationEngine = locationEngineDictionary.get(LocationEngine.Type.ANDROID);
+
```
Guardiola31337: NIT empty line

PR #123, opened 2018-04-18
Integrate Geofencing

electrostat:
Use Google's play-service geofencing solution to improve location data. Restrict to applications that Have Play-Services enabled.


libtelemetry/src/main/java/com/mapbox/android/telemetry/GeofenceJobService.java
```diff
@@ -0,0 +1,215 @@
...
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.List;
+
+import okhttp3.Call;
+import okhttp3.Callback;
+import okhttp3.Response;
+
+@RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+public class GeofenceJobService extends JobService implements Callback {
```
tobrun: could you explain to me why a JobService is needed for integrating geofences?

alfwatt:
Stale

Issue #125, opened 2018-04-24
Decouple Core and Telem Release Process


electrostat:
Decouple the release process for Telemetry and Core libraries. This allows them to be released independently and prevent any conflicts within the release process.

PR #131, opened 2018-05-01
Decouple Core and Telemetry Release

electrostat:
Create a process to easily decouple the release of Core and Telemetry libraries

fixes https://github.com/mapbox/mapbox-events-android/issues/125

### DIFF
--- a/gradle/mvn-push-android.gradle
+++ b/gradle/mvn-push-android.gradle
@@ -54,7 +54,6 @@ afterEvaluate { project ->
     uploadArchives {
         repositories {
             mavenDeployer {
-
                 beforeDeployment { MavenDeployment deployment -> signing.signPom(deployment) }
 
                 pom.groupId = GROUP


zugaldia:
Thanks for running with this. Could you post a quick comment explaining how the new process would work now (i.e. how to release one library not the other)?

electrostat:
@zugaldia 

As default, `OK_TO_RELEASE_CORE` and `OK_TO_RELEASE_TELEM` are hardcoded as `true` within the CircleCi config file. When you want to release only one of the libraries, we change the value of the opposite library (as in releasing telem w/o core, we change the `OK_TO_RELEASE_CORE` to false).  I tested this by releasing `SNAPSHOTS` from this branch. When done, just revert back before merging back to master.

This should be an additional step to check on our release checklist.

gradle/mvn-push-android.gradle
```diff
@@ -54,6 +54,7 @@ afterEvaluate { project ->
     uploadArchives {
         repositories {
             mavenDeployer {
+
```
zugaldia: This change seems accidental.

PR #144, opened 2018-05-14
Missing Application Context

electrostat:
Under certain circumstances, a LocationEvent will be created while `MapboxTelemetry.applicationContext` is null. Add in check and set correct context when creating event.

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
@@ -95,6 +95,7 @@ public class TelemetryService extends Service implements TelemetryCallback, Loca
 
   @Override
   public void onLocationChanged(Location location) {
+    checkApplicationContext();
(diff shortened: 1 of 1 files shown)


zugaldia:
Do we know the root issue for this to happen? Why is the Context null in the first place? Is this limited to a specific scenario (e.g. startup) or can it happen in other circumstances?

electrostat:
@zugaldia No we do not know the root issue or circumstances for this to happen, without example code for the crash. I did duplicate the crash by forcing the context to null earlier in the workflow.  It returns all the same errors and is the cause of the crash. So I added in a check and a fix for this specific issue. My guess is the cause is a weird background state that can arise under specific circumstances (i.e. Samsung).

zugaldia:
@electrostat Thanks for the additional information - in that case, let's go ahead with this workaround for now. Please cut a new ticket where we can research the root cause later on. I have the feeling that this could be a race condition during initialization but we'd need to run some tests with different set ups to confirm (and in that case, we might just want to drop the affected location update as it won't be a noticeable data issue).

PR #151, opened 2018-05-21
New Hashes and Blacklist

electrostat:
Add new hashes for certificates and create a blacklist system to ping and compare against when sending out data.

Fixes https://github.com/mapbox/mobile-telemetry/issues/18


electrostat:
Moving into Review Mode. 

CC'ing @drboyer for visibility on this.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -36,11 +39,40 @@ CertificatePinner provideCertificatePinnerFor(Environment environment) {
...
+
+  private Map<String, List<String>> removeBlacklistedPins(Map<String, List<String>> pins, Environment environment) {
+    String key = COM_EVENTS;
+
+    if (environment == Environment.CHINA) {
+      key = CHINA_EVENTS;
+    }
+
+    List<String> hashList = pins.get(key);
+    CertificateBlacklist certificateBlacklist = new CertificateBlacklist(MapboxTelemetry.applicationContext,
```
Guardiola31337: We're coupling the retrieve blacklist functionality to `CertificatePinnerFactory`. What about passing the blacklist somehow from `MapboxTelemetry` all the way down to `CertificatePinnerFactory`? Same idea as pins current implementation:

`MapboxTelemetry#createTelemetryClient` > `TelemetryClientFactory#obtainTelemetryClient` > `TelemetryClient` > `TelemetryClient#sendBatch` - `setting.getClient()` > `TelemetryClientSettings#configureHttpClient` > `CertificatePinnerFactory#provideCertificatePinnerFor`

IMO `CertificatePinnerFactory` shouldn't be responsible of retrieving the blacklist. Does that make sense? `

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,204 @@
...
+    File directory = context.getFilesDir();
+    ArrayList<String> blacklist = null;
+
+    if (directory.isDirectory()) {
+      File file = new File(directory, BLACKLIST_FILE);
+
+      if (file.exists()) {
+        try {
+          blacklist = getBlacklistContents(file);
+          blacklist.remove(0);
```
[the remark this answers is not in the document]
electrostat: Correct

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,204 @@
...
+      outputStream = context.openFileOutput(filename, Context.MODE_PRIVATE);
+      outputStream.write(fileContents.getBytes());
+      outputStream.close();
+    } catch (Exception exception) {
+      exception.printStackTrace();
+    }
+  }
+
+  @Override
+  public void onFailure(Call call, IOException exception) {
```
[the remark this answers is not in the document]
electrostat: correct. Goal was to not show this error to the developer. The system checks itself every day and will be updated then.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,204 @@
...
+    StringBuilder content = new StringBuilder("" + date.getTime() + "\n");
+
+    for (String key: revokedKeys) {
+      content.append(SHA256).append(key).append("\n");
+    }
+
+    return content.toString();
+  }
+
+  private static ArrayList<String> getBlacklistContents(final File file) throws IOException {
```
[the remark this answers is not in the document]
electrostat: Good catch. Was part of an older implementation accessing the method outside of this class.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
electrostat: What is the tangible benefit of this much more complicated routing? What are the cons of the current implementation?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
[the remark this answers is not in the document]
Guardiola31337: Great, could we reflect that in the code somehow?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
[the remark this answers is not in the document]
Guardiola31337: Got it. Could we add a comment like `// Intentionally empty` then?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,216 @@
...
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+class CertificateBlacklist implements Callback {
+  private static final String LOG_TAG = "MapboxBlacklist";
+  private final long DAY_IN_MILLISECONDS = 86400000;
+  private static final String BLACKLIST_FILE = "MapboxBlacklist";
+  private static final String SHA256 = "sha256/";
+  private static final String NEW_LINE = "\n/";
```
Guardiola31337: Typo? Shouldn't be `"\n"`?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -36,11 +39,38 @@ CertificatePinner provideCertificatePinnerFor(Environment environment) {
...
+      MapboxTelemetry.accessToken);
+
+    List blackList = certificateBlacklist.retrieveBlackList();
+
+    for (String hash: new ArrayList<>(hashList)) {
```
Guardiola31337: Creating an `ArrayList` here is no needed. We can use just `hashList`.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -61,13 +61,11 @@ private void addCertificatesPins(Map<String, List<String>> pins, CertificatePinn
     CertificateBlacklist certificateBlacklist = new CertificateBlacklist(MapboxTelemetry.applicationContext,
       MapboxTelemetry.accessToken);
 
-    ArrayList blackList = certificateBlacklist.retrieveBlackList();
+    List blackList = certificateBlacklist.retrieveBlackList();
 
-    if (blackList != null) {
```
Guardiola31337: Shouldn't we still check if the `blackList` is empty? We're traversing `hashList` even if `blackList` has no elements which I guess is not needed. Does that make sense?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
Guardiola31337: The main goal of the change is being able to test the new certificate blacklist functionality. Right now, we're not able to inject a `CertificateBlacklist` instance because we're creating a new instance of it within `CertificatePinnerFactory#removeBlacklistedPins`. We're calling `retrieveBlackList` which ends up retrieving the `File`, reading it and returning a `List` of its content which is not responsibility of `CertificatePinnerFactory`. Also, as commented in https://github.com/mapbox/mapbox-events-android/pull/151#discussion_r202375564 in order to access to the `accessToken` (needed for creating a `CertificateBlacklist`) we had to make it "globally" accessible in `MapboxTelemetry` which shouldn't be necessary.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -17,17 +17,21 @@
...
+  private CertificateBlacklist certificateBlacklist;
 
   /**
    * Based on http://square.github.io/okhttp/3.x/okhttp/okhttp3/CertificatePinner.html
    *
    * @return The CertificatePinner instance
    */
-  CertificatePinner provideCertificatePinnerFor(Environment environment) {
+  CertificatePinner provideCertificatePinnerFor(Environment environment, CertificateBlacklist certificateBlacklist) {
+    this.certificateBlacklist = certificateBlacklist;
```
Guardiola31337: This is not necessary. We can pass in the `CertificateBlacklist` through `addCertificatesPins` and then through `removeBlacklistedPins`.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,216 @@
...
+
+    saveBlackList(revokedKeys);
+  }
+
+  private String createListContent(List<String> revokedKeys) {
+    Date date = new Date();
+
+    StringBuilder content = new StringBuilder(date.getTime() + NEW_LINE);
+
+    for (String key: revokedKeys) {
```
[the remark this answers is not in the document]
electrostat: Need more specifics on this. What are you looking for here?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,202 @@
...
+import okhttp3.Request;
+import okhttp3.Response;
+
+class CertificateBlacklist implements Callback {
+  private static final String LOG_TAG = "MapboxBlacklist";
+  private static final String BLACKLIST_FILE = "MapboxBlacklist";
+  private static final String SHA256 = "sha256/";
+  private static final String NEW_LINE = "\n";
+  private static final String QUESTION_MARK = "?";
+  private static final String EQAL_SIGN = "=";
```
Guardiola31337: Typo `*EQAL`

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,202 @@
...
+  private static final int BLACKLIST_HEAD = 0;
+  private static final String REVOKED_CERT_KEYS = "RevokedCertKeys";
```
Guardiola31337: > Private field 'REVOKED_CERT_KEYS' is never used

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,202 @@
...
+  private static final String COM_CONFIG_ENDPOINT = "https://api.mapbox.com/events-config";
+  private static final String CHINA_CONFIG_ENDPOINT = "https://api.mapbox.cn/events-config";
+  private static final String ACCESS_TOKEN_QUERY_PARAMETER = "access_token";
+  private static final String EXTRACTION_EXEMPTION = "Blacklist Response Extraction Failed";
```
Guardiola31337: > Private field 'EXTRACTION_EXEMPTION' is never used

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,202 @@
...
+        Log.e(LOG_TAG, RETRIEVE_TIME_FAIL, exception);
+      }
+    }
+
+    return lastUpdateTime;
+  }
+
+  void updateBlacklist() {
+    Request request = new Request.Builder()
+      .url(determineConfigEndpoint() + QUESTION_MARK + ACCESS_TOKEN_QUERY_PARAMETER + EQAL_SIGN + accessToken)
```
Guardiola31337: As commented in https://github.com/mapbox/mapbox-events-android/pull/151#discussion_r202348319 what about 👇

> passing a `HttpUrl` instead (we're currently doing that in `TelemetryClient` for example)? That way we'd able to add query parameters easily 👀 https://square.github.io/okhttp/3.x/okhttp/okhttp3/HttpUrl.Builder.html#addQueryParameter-java.lang.String-java.lang.String-

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,202 @@
...
+
+    if (directory.isDirectory()) {
+      File file = new File(directory, BLACKLIST_FILE);
+
+      if (file.exists()) {
+        try {
+          blacklist = obtainBlacklistContents(file);
+          blacklist.remove(BLACKLIST_HEAD);
+        } catch (IOException exception) {
+          Log.e(LOG_TAG, RETRIEVE_BLACKLIST_FAIL, exception);
```
Guardiola31337: > Using 'Log' instead of 'Timber'

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,202 @@
...
+    File file = new File(directory, BLACKLIST_FILE);
+
+    long lastUpdateTime = 0;
+
+    if (file.exists()) {
+      try {
+        List<String> blacklist = obtainBlacklistContents(file);
+        lastUpdateTime = Long.valueOf(blacklist.get(BLACKLIST_HEAD));
+      } catch (IOException exception) {
+        Log.e(LOG_TAG, RETRIEVE_TIME_FAIL, exception);
```
Guardiola31337: > Using 'Log' instead of 'Timber'

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,202 @@
...
+  private void saveBlackList(List<String> revokedKeys) {
+    String fileContents = createListContent(revokedKeys);
+    FileOutputStream outputStream;
+
+    try {
+      outputStream = context.openFileOutput(BLACKLIST_FILE, Context.MODE_PRIVATE);
+      outputStream.write(fileContents.getBytes());
+      outputStream.close();
+    } catch (Exception exception) {
+      Log.e(LOG_TAG, SAVE_BLACKLIST_FAIL, exception);
```
Guardiola31337: > Using 'Log' instead of 'Timber'

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  @Override
+  public void onFailure(Call call, IOException exception) {
+    Log.e(LOG_TAG, REQUEST_FAIL, exception);
```
Guardiola31337: > Using 'Log' instead of 'Timber'

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,202 @@
...
+    try {
+      ApplicationInfo appInformation = context.getPackageManager().getApplicationInfo(context.getPackageName(),
+        PackageManager.GET_META_DATA);
+
+      if (appInformation != null && appInformation.metaData != null) {
+        serverInformation = setupChain.obtainServerInformation(appInformation.metaData);
+        return ENDPOINTS.get(serverInformation.getEnvironment());
+      }
+    } catch (Exception exception) {
+      Log.e(LOG_TAG, ENDPOINT_DETERMINATION_FAIL, exception);
```
Guardiola31337: > Using 'Log' instead of 'Timber'

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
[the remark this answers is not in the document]
Guardiola31337: As commented in https://github.com/mapbox/mapbox-events-android/pull/151#discussion_r202349738

> NIT - bad formatting
> ```java
> for (String key : revokedKeys) {
> ```

`⌥⌘L` is your friend 😉

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    return null;
```
Guardiola31337: `context.getPackageManager().getApplicationInfo` could throw a `PackageManager.NameNotFoundException` - this shouldn't happen but if it does we'd return `null` which would cause a crash in the call site https://github.com/mapbox/mapbox-events-android/blob/f4bba88eda2949f5d1b6d88592ca2e290efc6be8/libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java#L108

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -36,11 +38,39 @@ CertificatePinner provideCertificatePinnerFor(Environment environment) {
...
+    pins = removeBlacklistedPins(pins, environment, certificateBlacklist);
+
     for (Map.Entry<String, List<String>> entry : pins.entrySet()) {
       for (String pin : entry.getValue()) {
         builder.add(entry.getKey(), pin);
       }
     }
   }
+
+  private Map<String, List<String>> removeBlacklistedPins(Map<String, List<String>> pins, Environment environment,
```
Guardiola31337: NIT - some unnecessary new lines

libtelemetry/src/main/java/com/mapbox/android/telemetry/RevokedKeys.java
```diff
@@ -0,0 +1,12 @@
+package com.mapbox.android.telemetry;
+
+import java.util.List;
+
+class RevokedKeys {
+
+  private List<String> RevokedCertKeys;
```
Guardiola31337: 👀 Java variables naming convention https://docs.oracle.com/javase/tutorial/java/nutsandbolts/variables.html

> If the name you choose consists of only one word, spell that word in all lowercase letters. If it consists of more than one word, capitalize the first letter of each subsequent word.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -36,11 +38,39 @@ CertificatePinner provideCertificatePinnerFor(Environment environment) {
...
+                                                          CertificateBlacklist certificateBlacklist) {
+    String key = COM_EVENTS;
+
+    if (environment == Environment.CHINA) {
+      key = CHINA_EVENTS;
+    }
+
+    List<String> hashList = pins.get(key);
```
Guardiola31337: At this point we could obtain the key from the `pins` as follows:

```java
Set<String> pinsKey = pins.keySet();
String key = pinsKey.iterator().next();
```

We can guarantee that the `pins` instance has only one entry so we don't need to check for the key explicitly - `Environment` could be removed all the way up and `COM_EVENTS` and  `CHINA_EVENTS` constants too.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,220 @@
...
+    {
+      put(Environment.COM, COM_CONFIG_ENDPOINT);
+      put(Environment.CHINA, CHINA_CONFIG_ENDPOINT);
+    }
+  };
+  private Context context;
+  private String accessToken;
+
+  CertificateBlacklist(Context context, String accessToken) {
+    this.context = context;
```
andrlee: I wonder if both context and accessToken variables should be immutable?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,220 @@
...
+    this.accessToken = accessToken;
+  }
+
+  List<String> retrieveBlackList() {
```
andrlee: What are some ways to make this method content more readable? One potential improvement i see here: have less nested levels of conditions.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,220 @@
...
+    File directory = context.getFilesDir();
+    List<String> blacklist = new ArrayList<>();
+
+    if (directory.isDirectory()) {
+      File file = new File(directory, BLACKLIST_FILE);
+
+      if (file.exists()) {
+        try {
```
andrlee: I wonder if we can use "try with resources" here, which should significantly simplify error handling logic.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+      }
+    }
+
+    return blacklist;
+  }
+
+  boolean daySinceLastUpdate() {
+    long millisecondDiff = System.currentTimeMillis() - retrieveLastUpdateTime();
+    long dayInMilliseconds = 86400000;
```
andrlee: I wonder why this value is hardcoded and if there are any downsides to hardcoding it.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  private long retrieveLastUpdateTime() {
+    File directory = context.getFilesDir();
+    File file = new File(directory, BLACKLIST_FILE);
+
+    long lastUpdateTime = 0;
+
+    if (file.exists()) {
+      try {
```
andrlee: I wonder if we can use "try with resources" here as well.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,220 @@
...
+    long dayInMilliseconds = 86400000;
+
+    return millisecondDiff >= dayInMilliseconds;
+  }
+
+  private long retrieveLastUpdateTime() {
+    File directory = context.getFilesDir();
+    File file = new File(directory, BLACKLIST_FILE);
+
+    long lastUpdateTime = 0;
```
andrlee: Wouldn't it be a redundant initialization?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,220 @@
...
+    OkHttpClient client = new OkHttpClient();
+    client.newCall(request).enqueue(this);
+  }
+
+  private void saveBlackList(List<String> revokedKeys) {
+    String fileContents = createListContent(revokedKeys);
+    FileOutputStream outputStream;
+
+    try {
+      outputStream = context.openFileOutput(BLACKLIST_FILE, Context.MODE_PRIVATE);
```
andrlee: I wonder if we can use "try with resources" here as well.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -36,11 +37,36 @@ CertificatePinner provideCertificatePinnerFor(Environment environment) {
...
   }
+
+  private Map<String, List<String>> removeBlacklistedPins(Map<String, List<String>> pins,
+                                                          CertificateBlacklist certificateBlacklist) {
+    Set<String> pinsKey = pins.keySet();
+    String key = pinsKey.iterator().next();
+
+    List<String> hashList = pins.get(key);
+
+    if (certificateBlacklist != null) {
```
andrlee: What are some ways to make this block of code more readable?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -36,11 +37,34 @@ CertificatePinner provideCertificatePinnerFor(Environment environment) {
...
   }
+
+  private Map<String, List<String>> removeBlacklistedPins(Map<String, List<String>> pins,
+                                                          CertificateBlacklist certificateBlacklist) {
+    Set<String> pinsKey = pins.keySet();
+    String key = pinsKey.iterator().next();
+
+    List<String> hashList = pins.get(key);
+
+    List blackList = certificateBlacklist.retrieveBlackList();
```
andrlee: What are some ways to make this block of code more readable?

libtelemetry/src/main/java/com/mapbox/android/telemetry/ChinaCertificatePins.java
```diff
@@ -26,6 +26,57 @@
           add("sha256/+O+QJCmvoB/FkTd0/5FvmMSvFbMqjYU+Txrw1lyGkUQ=");
           add("sha256/zUIraRNo+4JoAYA7ROeWjARtIoN4rIEbCpfCRQT6N6A=");
           add("sha256/r/mIkG3eEpVdm+u/ko/cwxzOMo1bk4TyHIlByibiA5E=");
+
+          add("sha256/T4XyKSRwZ5icOqGmJUXiDYGa+SaXKTGQXZwhqpwNTEo=");
```
andrlee: Are there any security implications to hardcoding certificate pins?

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -55,12 +55,14 @@
...
   private PermissionCheckRunnable permissionCheckRunnable = null;
   private CopyOnWriteArraySet<TelemetryListener> telemetryListeners = null;
+  private CertificateBlacklist certificateBlacklist;
   static Context applicationContext = null;
 
   public MapboxTelemetry(Context context, String accessToken, String userAgent) {
     initializeContext(context);
     initializeQueue();
     checkRequiredParameters(accessToken, userAgent);
+    checkBlacklist();
```
andrlee: Wouldn't it make more sense to pass context and accessToken as parameters to this method to be consistent with other initialization methods?

libtelemetry/src/main/java/com/mapbox/android/telemetry/RevokedKeys.java
```diff
@@ -0,0 +1,12 @@
+package com.mapbox.android.telemetry;
+
+import java.util.List;
+
+class RevokedKeys {
```
andrlee: What are some benefits and side-effects of having this class defined? It looks like it just simply wraps the list of keys.

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClient.java
```diff
@@ -24,16 +24,19 @@
   private static final String EXTRA_DEBUGGING_LOG = "Sending POST to %s with %d event(s) (user agent: %s) "
     + "with payload: %s";
 
-  private String accessToken = null;
-  private String userAgent = null;
+  private String accessToken;
```
andrlee: I like how these member variables no longer have redundant initializations but could we also make them immutable?

electrostat: Both can be updated, so making them `final` would require removing that functionality.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,213 @@
...
+  private List<String> obtainBlacklistContents(File file) throws IOException {
+    InputStream inputStream = new FileInputStream(file);
+    BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream));
+
+    List<String> blacklist = new ArrayList<>();
+
+    boolean done = false;
+
+    while (!done) {
+      String line = reader.readLine();
```
andrlee: reader.readLine() might throw an IOException, then 
`reader.close();
  inputStream.close();`
block will not be called on line 175 and 176. right? shouldn't we handle IOException internally and call close() in finally block? one option is to return an empty list if any of the I/O operations fail. thoughts?

libtelemetry/src/main/java/com/mapbox/android/telemetry/ChinaCertificatePins.java
electrostat: Great question. I am not too sure. This system does include a blacklist to turn off any compromised hashes and we have always had the pins hardcoded (see above this point in the code), but something we should discuss with the security team and talk about any solutions if there are problems. I'll ping them.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
electrostat: No may not be set within the if-statement. Can't return an uninitialized variable.

libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/TelemetryServiceTest.java
```diff
@@ -174,8 +174,8 @@ private void stopService(Intent serviceIntent, TelemetryService[] telemetryServi
 
   private void waitUntilServiceIsDestroyed() throws InterruptedException {
     // Have to wait a bit until the system calls Service#onDestroy()
-    // Note that it only needs 10 milliseconds to clean the resources up
-    Thread.sleep(10);
+    // Note that it only needs 15 milliseconds to clean the resources up
+    Thread.sleep(15);
```
andrlee: How did we come up with the 15 milliseconds threshold?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
andrlee: If i remember correctly, in java a local variable of long type is initialized to 0L by default. could you please check with the java spec?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,225 @@
...
+        done = (line == null);
+
+        if (line != null && !line.isEmpty()) {
+          blacklist.add(line);
+        }
+      }
+    } catch (Exception exception) {
+      logger.error(READLINE_FAIL, exception.getMessage());
+    } finally {
+      reader.close();
```
andrlee: Actually, both close() methods on line 184 and 185 can potentially throw an IOException, hence they both should be enclosed into try-catch block.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,225 @@
...
+    ServerInformation serverInformation;
+    try {
+      ApplicationInfo appInformation = context.getPackageManager().getApplicationInfo(context.getPackageName(),
+        PackageManager.GET_META_DATA);
+
+      if (appInformation != null && appInformation.metaData != null) {
+        serverInformation = setupChain.obtainServerInformation(appInformation.metaData);
+        return ENDPOINTS.get(serverInformation.getEnvironment());
+      }
+    } catch (Exception exception) {
```
andrlee: I wonder if it is possible not to use catch all exception here and be a little more explicit about error handling?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -36,11 +37,37 @@ CertificatePinner provideCertificatePinnerFor(Environment environment) {
     return CERTIFICATES_PINS.get(environment);
   }
 
-  private void addCertificatesPins(Map<String, List<String>> pins, CertificatePinner.Builder builder) {
+  private void addCertificatesPins(Map<String, List<String>> pins, CertificatePinner.Builder builder,
+                                   CertificateBlacklist certificateBlacklist) {
+    pins = removeBlacklistedPins(pins, certificateBlacklist);
```
andrlee: According to the name of this method it should only add pins to the blacklist, now it's actually also does cleanup, hence would it make sense to decouple cleanup logic and call removeBlacklistedPins on line 30 just right before addCertificatesPins is called?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -36,11 +37,37 @@ CertificatePinner provideCertificatePinnerFor(Environment environment) {
...
   }
+
+  private Map<String, List<String>> removeBlacklistedPins(Map<String, List<String>> pins,
+                                                          CertificateBlacklist certificateBlacklist) {
+    Set<String> pinsKey = pins.keySet();
+    String key = pinsKey.iterator().next();
+
+    List<String> hashList = pins.get(key);
+
+    List blackList = certificateBlacklist.retrieveBlackList();
```
andrlee: Shouldn't the block of code between line 58 and 62 be called first thing within this method?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -36,11 +37,37 @@ CertificatePinner provideCertificatePinnerFor(Environment environment) {
...
     for (Map.Entry<String, List<String>> entry : pins.entrySet()) {
       for (String pin : entry.getValue()) {
         builder.add(entry.getKey(), pin);
       }
     }
   }
+
+  private Map<String, List<String>> removeBlacklistedPins(Map<String, List<String>> pins,
+                                                          CertificateBlacklist certificateBlacklist) {
+    Set<String> pinsKey = pins.keySet();
```
andrlee: Should line 53-56 be decoupled into a helper method?

libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/TelemetryServiceTest.java
electrostat: Intermittent failure of our `checksLocationReceiverIsDownWhenServiceStopped` test. Has been an issue for awhile. Pushed milliseconds to 15 to see if that would affect the test and it stopped failing. Probably can play with the value more, if desired.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  private void saveBlackList(List<String> revokedKeys) {
+    String fileContents = createListContent(revokedKeys);
+    FileOutputStream outputStream;
+
+    try {
+      outputStream = context.openFileOutput(BLACKLIST_FILE, Context.MODE_PRIVATE);
+      outputStream.write(fileContents.getBytes());
+      outputStream.close();
```
andrlee: what if write() on line 134 throws? shouldn't close() be called in finally block?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,226 @@
...
+      while (!done) {
+        String line = reader.readLine();
+        done = (line == null);
+
+        if (line != null && !line.isEmpty()) {
+          blacklist.add(line);
+        }
+      }
+
+      reader.close();
```
andrlee: if readLine() on line 174 thows then close() on line 182 and 183 will not be closed. should we call close() in finally block?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,243 @@
...
+      while (!done) {
+        String line = reader.readLine();
+        done = (line == null);
+
+        if (line != null && !line.isEmpty()) {
+          blacklist.add(line);
+        }
+      }
+
+    } catch (Exception exception) {
```
andrlee: catch (IOException)

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,243 @@
...
+      logger.error(READLINE_FAIL, exception.getMessage());
+    } finally {
+      try {
+        inputStream.close();
+      } catch (Exception exception) {
```
andrlee: catch (IOException)

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,243 @@
...
+        logger.error(CLOSE_STREAM_FAIL, exception.getMessage());
+      }
+
+      try {
+        reader.close();
+      } catch (Exception exception) {
```
andrlee: catch (IOException)

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -43,4 +45,36 @@ private void addCertificatesPins(Map<String, List<String>> pins, CertificatePinn
       }
     }
   }
+
+  private Map<String, List<String>> removeBlacklistedPins(Map<String, List<String>> pins,
+                                                          CertificateBlacklist certificateBlacklist) {
+    List blackList = certificateBlacklist.retrieveBlackList();
```
andrlee: should list be opaque here?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  private void saveBlackList(List<String> revokedKeys) {
+    String fileContents = createListContent(revokedKeys);
+    FileOutputStream outputStream = null;
+
+    try {
+      outputStream = context.openFileOutput(BLACKLIST_FILE, Context.MODE_PRIVATE);
+      outputStream.write(fileContents.getBytes());
+    } catch (Exception exception) {
```
andrlee: what about this block? catch (IOException exception)

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -0,0 +1,243 @@
...
+      logger.error(SAVE_BLACKLIST_FAIL, exception.getMessage());
+    } finally {
+      try {
+        outputStream.close();
+      } catch (Exception exception) {
```
andrlee: catch (IOException exception)

PR #155, opened 2018-05-25
Make location event constructor package private

Guardiola31337:
- Makes `LocationEvent` constructor package-private


zugaldia:
Isn't this considered a SEMVER change?

Guardiola31337:
> Isn't this considered a SEMVER change?

Nah, I'd say it was a bug fix. `LocationEvent` was only used internally.

Issue #147, opened 2018-05-17
TelemetryLocationEnabler.updateTelemetryLocationState  NPE issue


lilykaiser:
Copying from @kingfisherphuoc https://github.com/mapbox/mapbox-gl-native/issues/11929

**Platform:**
**Mapbox SDK version:**

### Steps to trigger behavior

 1. open map
 2. Using it while GPS is moving (Using Lockito or some mocklocation app)
 3. Get error as below:

```
Caused by java.lang.NullPointerException
Attempt to invoke virtual method 'com.mapbox.android.telemetry.TelemetryLocationEnabler$LocationState com.mapbox.android.telemetry.TelemetryLocationEnabler.updateTelemetryLocationState(com.mapbox.android.telemetry.TelemetryLocationEnabler$LocationState)' on a null object reference
arrow_right
com.mapbox.android.telemetry.TelemetryService.disableTelemetryLocationState (TelemetryService.java:240)
com.mapbox.android.telemetry.TelemetryService.onDestroy (TelemetryService.java:64)
android.app.ActivityThread.handleStopService (ActivityThread.java:3771)
android.app.ActivityThread.-wrap30 (ActivityThread.java)
android.app.ActivityThread$H.handleMessage (ActivityThread.java:1752)
android.os.Handler.dispatchMessage (Handler.java:102)
android.os.Looper.loop (Looper.java:154)
android.app.ActivityThread.main (ActivityThread.java:6776)
java.lang.reflect.Method.invoke (Method.java)
com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1518)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1408)
```

### Expected behavior
We should not get this issue.
### Actual behavior
App crashed.

Guardiola31337:
Hey @kingfisherphuoc 👋 thanks for reaching out and report your issue.

Could you add some more information for reproducing the crash (minimal example) or any other specifics around your test setup and some information on used device? What version of the Maps SDK are you using? If there is any sample or pseudo code you can provide, that would be also really helpful.

electrostat:
I follow the logic. This looks like it'll fix the issue. I'll work on a solution.

lilykaiser:
Fixed in https://github.com/mapbox/mapbox-events-android/pull/156

PR #156, opened 2018-05-25
TelemetryLocationEnabler Null Bug 

electrostat:
Under certain circumstances, `TelemetryLocationEnabler` is null. Add in check to initialize, before being used.

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
@@ -205,7 +205,7 @@ public class TelemetryService extends Service implements TelemetryCallback, Loca
     isLocationEnablerFromPreferences = intent.getBooleanExtra(IS_LOCATION_ENABLER_FROM_PREFERENCES, true);
 
     if (isLocationEnablerFromPreferences) {
-      telemetryLocationEnabler = new TelemetryLocationEnabler(true);
+      checkLocationEnabler();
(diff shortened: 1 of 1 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
```diff
@@ -239,6 +239,10 @@ private void unregisterTelemetryReceiver() {
 
   private void disableTelemetryLocationState() {
     if (isLocationEnablerFromPreferences) {
+      if (telemetryLocationEnabler == null) {
+        telemetryLocationEnabler = new TelemetryLocationEnabler(true);
```
Guardiola31337: Could you extract a _Lazy initialization_ method as suggested here https://github.com/mapbox/mapbox-events-android/issues/147#issuecomment-392010793?
> We could extract a _Lazy initialization_ method and use it accordingly in both call sites (when enabling and disabling).

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
```diff
@@ -253,6 +250,12 @@ private void checkApplicationContext() {
     }
   }
 
+  private void checkLocationEnabler() {
```
Guardiola31337: As we're creating a new instance of `TelemetryLocationEnabler` if not already created, what about calling this method `createLocationEnabler` instead?

PR #157, opened 2018-05-30
Oreo Background Service Crash

electrostat:
With developers targeting Api 26+ now, background service no longer works and is creating a crash. Reworking system to prevent this crash and be Oreo optimized.


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -488,7 +490,12 @@ private PermissionCheckRunnable obtainPermissionCheckRunnable() {
   }
 
   private void startLocation() {
-    applicationContext.startService(obtainLocationServiceIntent());
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+      applicationContext.startForegroundService(obtainLocationServiceIntent());
```
LukasPaczos: You could additionally ensure that the service is not started from the background by wrapping this line with `ProcessLifecycleOwner.get().lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)`.

electrostat: Nice catch.

florianPOLARSTEPS: starting a long- lived foreground service should only be done when absolutely necessary. It will trigger the OS (Android 8) to show a notification "Application X" is consuming your battery.

electrostat: @florianPOLARSTEPS that is correct, but I am killing the service when the app enters the background to avoid this issue. When returning to the foreground, the service is spun up again and continues collecting and processing event data.

LukasPaczos:
What's the advantage of using a foreground service in this context? Couldn't we use a background, sticky service? How sensitive is the event processing that it can't be interrupted?

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
florianPOLARSTEPS: @electrostat Alright, thanks... I just noticed the LifecyleObserver, sorry about that.

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
```diff
@@ -254,4 +260,16 @@ TelemetryService obtainService() {
       return TelemetryService.this;
     }
   }
+
+  private void initiateForegroundService() {
+    Notification notification = new Notification();
+    startForeground(1375, notification);
```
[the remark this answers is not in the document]
electrostat: Per the above comment, I actually disable the foreground service when the application enters the background, thus removing the notification.

The Android P permission is a new one to me. Haven't looked at the documentation for that yet, but will give it a look. I'm planning on being `P` compliant much earlier than we are getting to `Oreo`.

electrostat:
@LukasPaczos Thanks for bringing up this line of thought. I did some digging and realized I made a mistake in my interpretation of the documentation. I read it as removing the `startService` functionality all together, not just in the background. So they were forcing everyone over to using a foreground service. That's incorrect. We can't `start` a service while in the background through this method and need to use startForeground service instead. Given this new insight, your earlier suggestion is likely all we need to fix this issue.

Our system is robust and can be interrupted freely, so no worries there. My original solution would've faced the same issues.

I will make the change and test accordingly.

LukasPaczos:
I haven't tested it, but based on the Maps SDK setup if a user calls `Mapbox#getInstance` (effectively initializing telemetry) from the `Application#onCreate`, or `#onCreate` of the first Activity (based on [ProcessLifecycleOwner docs](https://developer.android.com/reference/android/arch/lifecycle/ProcessLifecycleOwner)) the app's process won't be in a `Lifecycle.State.STARTED` state yet. This means, that we need to keep lifecycle components around to start the service when we eventually do reach that state on Oreo devices, is that right or am I missing something here?

gradle/dependencies.gradle
```diff
@@ -2,22 +2,23 @@ ext {
...
+            gmsLocation         : '11.0.4',
+            junit               : '4.12',
+            supportLibVersion   : '26.1.0',
+            constraintLayout    : '1.0.2',
+            mockito             : '2.11.0',
+            testRunnerVersion   : '1.0.1',
+            okhttp3             : '3.10.0',
+            gson                : '2.8.2',
+            espressoVersion     : '3.0.1',
+            archLifecycleVersion: "1.0.0"
```
LukasPaczos: Would there be any benefits to bumping arch components version? I think `1.1.1` requires support lib 27+, but maybe we could use `1.1.0`?

electrostat: Nope 1.1.1 is good with `supportLib` of 26.1.0

gradle/dependencies.gradle
```diff
@@ -33,6 +34,10 @@ ext {
             // play services
             gmsLocation            : "com.google.android.gms:play-services-location:${version.gmsLocation}",
 
+            //architecture components
+            archLifecycleExtensions: "android.arch.lifecycle:extensions:${version.archLifecycleVersion}",
```
LukasPaczos: We can save some weight here by using `android.arch.lifecycle:runtime:${version.archLifecycleVersion}` dependency instead of `extensions` as we are not using `ViewModel` or `LiveData`.

LukasPaczos: Capturing from a chat that it seems like `runtime` does not contain `ProcessLifecycleOwner` so we need to stick to `extensions`.

andrlee: @electrostat any good reason why @LukasPaczos suggestion to cut on weight was not addressed and PR still got merged?

electrostat: @andrlee  Yes, per the above comment we discussed that `ProcessLifecycleOwner' is not included in `runtime`, so we had to keep `extensions`. We tested and discussed over slack the issue and was just captures in the brief above comment.

andrlee: we could still exclude `livedata` and `viewmodel` right? We shouldn't be including unnecessary deps.

PR #160, opened 2018-06-05
Schema Tests

electrostat:
Create tests that compare local event classes to our api schema for those events. Makes sure we do not release events that do not match the schema.


electrostat:
After pulling down the schema and running tests off of it, I have noticed that we are missing events and parameters of other events. None of the tests are passing at this point, because none of the events meet spec. Below are a list of test changes and event changes that need to be done to progress to spec.

- all events are missing `userAgent`
- any event that uses a `number` will need to have a separate method to test this against `doubles`, `floats`, etc...
- `LocationEvent` has horizontal accuracy, needs to be removed
- navigation events need to have separate test path to handle metadata 
- navigation events are missing `received` and `token` are these server side designators or do they need to be included when sending events back?
- Missing Events (client-side): `OverlappingAudio`, `SignificantSpeedDifference`, `Turnstile`, and `Visit` - which do we need, what is timetable for having these events integrated?

Can you help answer above questions?
cc: @zugaldia @sadiew @mapbox/navigation-android

libtelemetry/src/test/java/com/mapbox/android/telemetry/SchemaTest.java
```diff
@@ -0,0 +1,528 @@
...
+
+  @Test
+  public void checkAppUserTurnstileEvent() throws Exception {
+    JsonObject schema = grabSchema(APP_USER_TURNSTILE);
+    List<Field> fields = grabClassFields(AppUserTurnstile.class);
+
+    System.out.println("schema: " + schema);
+    System.out.println("fields: " + fields);
+
+    assertEquals(schema.size(), fields.size());
```
Guardiola31337: Generally is preferable having only 1 assertion per test 👀 https://speakerdeck.com/guardiola31337/elegant-unit-testing-mobilization-2016?slide=57
If this tests fails we won't know if it was because of an error with the fields size or because the `schemaContainsFields` assertion. What about creating different tests for each check?
The same thing happens with the rest of the tests 👇

libtelemetry/src/test/java/com/mapbox/android/telemetry/SchemaTest.java
```diff
@@ -0,0 +1,528 @@
...
+      System.out.println(typePrimitive + ": " + jsonElement);
+      assertTrue(typePrimitive.equals(jsonElement));
+    } else {
+      JsonArray arrayOfTypes = schema.getAsJsonArray("type");
+      System.out.println(arrayOfTypes + ": " + jsonElement);
+      assertTrue(arrayOfTypes.contains(jsonElement));
+    }
+  }
+
+  private void unpackSchemas(Response responseData) throws IOException, JSONException {
```
Guardiola31337: > Exception 'org.json.JSONException' is never thrown in the method

libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationEvent.java
```diff
@@ -30,8 +30,6 @@
   private String operatingSystem;
   @SerializedName("applicationState")
   private String applicationState;
-  @SerializedName("horizontalAccuracy")
-  private Float accuracy = null;
```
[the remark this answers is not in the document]
electrostat: That is indeed to spec, just checked. We can wrap around and talk to the data team, but I recall that they do not use that variable when using location data.

libtelemetry/src/test/java/com/mapbox/android/telemetry/SchemaTest.java
```diff
@@ -0,0 +1,528 @@
...
+  private static final String NAVIGATION_CANCEL = "navigation.cancel";
+  private static final String NAVIGATION_DEPART = "navigation.depart";
+  private static final String NAVIGATION_FASTER_ROUTE = "navigation.fasterRoute";
+  private static final String NAVIGATION_FEEDBACK = "navigation.feedback";
+  private static final String NAVIGATION_REROUTE = "navigation.reroute";
+  private ArrayList<JsonObject> schemaArray;
+
+  @Before
+  public void downloadSchema() throws Exception {
+    unpackSchemas();
```
Guardiola31337: We're still unzipping the schema on every test which is not necessary 👀 https://github.com/mapbox/mapbox-events-android/pull/160#discussion_r202411192

Although I'd personally rather creating a new task as suggested ☝️ we could also use the [`@BeforeClass`](http://junit.sourceforge.net/javadoc/org/junit/BeforeClass.html) annotation here 👀 

> Annotating a `public static void` no-arg method with `@BeforeClass` causes it to be run once before any of the test methods in the class.

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
```diff
@@ -45,17 +45,25 @@
   private Boolean batteryPluggedIn;
   private Integer batteryLevel;
   private String connectivity;
+  private String tripIdentifier;
+  private Integer legIndex;
+  private Integer legCount;
+  private Integer stepIndex;
+  private Integer voiceIndex = null;
```
andrlee: What are benefits of having this value initialized to null?

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
```diff
@@ -45,17 +45,25 @@
...
+  private Integer bannerIndex = null;
+  private Integer totalStepCount;
 
   public NavigationMetadata(Date startTimestamp, int distanceCompleted, int distanceRemaining, int durationRemaining,
```
andrlee: It looks like the constructor became quite unmaintainable. Any thoughts on how we can improve on readability? Perhaps consider Builders?

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
```diff
@@ -45,17 +45,25 @@
   private Boolean batteryPluggedIn;
   private Integer batteryLevel;
   private String connectivity;
+  private String tripIdentifier;
+  private Integer legIndex;
```
andrlee: What are some benefits of using boxed type here as opposed to primitive?

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
```diff
@@ -320,6 +365,14 @@ private NavigationMetadata(Parcel in) {
     batteryPluggedIn = in.readByte() != 0x00;
     batteryLevel = in.readInt();
     connectivity = in.readString();
+    tripIdentifier = in.readString();
+    legIndex = in.readInt();
```
andrlee: i assume readInt() returns primitive, would this assignment have a performance implication?

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
electrostat: This is helpful for the null checks when serializing the data lower in the class. We want to know what to write in the bytes.

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
electrostat: If I remember correctly, we purposefully moved away from Builders when we came up with the new implementation at 3.0.0. Our goal was to benefit developers. The good news is this can be moved out of the sdk during the refactor and external sdks can change it up or handle this as they want to.

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
electrostat: I see what happened here. While attempting to keep consistency, I mimicked `Integer` we use for the `null` checks when serializing. Will remove those that do not require the null check.

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
electrostat: see above

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
andrlee: With that huge number of mandatory/optional parameters should we consider reverting back to builders? public constructors with many parameters of the same type are super error prone.

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
andrlee: i'd be interested to evaluate an option of replacing boxed types with Optional<T> to model optional parameters.

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
```diff
@@ -35,27 +35,35 @@
...
+  private Integer voiceIndex = null;
+  private Integer bannerIndex = null;
+  private int totalStepCount;
 
   public NavigationMetadata(Date startTimestamp, int distanceCompleted, int distanceRemaining, int durationRemaining,
-                            String sdKIdentifier, String sdkVersion, int eventVersion, String sessionIdentifier,
+                            String sdkIdentifier, String sdkVersion, int eventVersion, String sessionIdentifier,
                             double lat, double lng, String geometry, String profile, boolean isSimulation,
-                            String locationEngine, int absoluteDistanceToDestination) {
+                            String locationEngine, int absoluteDistanceToDestination, String tripIdentifier,
```
Guardiola31337: This broke SemVer and should have been released as a major version (was released as patch `3.2.1`) or backporting the old logic and deprecating the old/to be removed constructor. We've noticed when Testing Maps SDK `6.6.0` releases downstream 👀 https://github.com/mapbox/mapbox-navigation-android/pull/1310 cc @andrlee @electrostat

libtelemetry/src/main/java/com/mapbox/android/telemetry/FeedbackEventData.java
```diff
@@ -35,14 +34,6 @@ String getSource() {
     return source;
   }
 
-  String getAudio() {
-    return audio;
-  }
-
-  public void setAudio(String audio) {
```
Guardiola31337: Same here a `public` method was removed which means breaking SemVer.

cc @andrlee @electrostat

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
```diff
@@ -209,14 +223,6 @@ String getAudioType() {
     return audioType;
   }
 
-  Integer getStepCount() {
-    return stepCount;
-  }
-
-  public void setStepCount(Integer stepCount) {
```
Guardiola31337: Same here as well. A `public` method was removed which means breaking SemVer.

cc @andrlee @electrostat

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationMetadata.java
zugaldia: Good catch @Guardiola31337. At a first glance this seems to only affect the Nav SDK but cc'ing @tobrun here to confirm the Maps SDK side.

PR #171, opened 2018-07-02
Update Play Services Location to 15.0.1

electrostat:
Update `Play-Services-Location` from 11.0.4 to 15.0.1

### DIFF
--- a/liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
+++ b/liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
@@ -1,5 +1,6 @@
 package com.mapbox.android.core.location;
 
+import android.annotation.SuppressLint;
 import android.content.Context;
 import android.location.Location;
 import android.os.Bundle;
@@ -8,22 +9,27 @@ import android.support.annotation.Nullable;
 
 import com.google.android.gms.common.ConnectionResult;
 import com.google.android.gms.common.api.GoogleApiClient;
-import com.google.android.gms.location.LocationListener;
+import com.google.android.gms.location.FusedLocationProviderClient;
+import com.google.android.gms.location.LocationCallback;
 import com.google.android.gms.location.LocationRequest;
+import com.google.android.gms.location.LocationResult;
 import com.google.android.gms.location.LocationServices;
+import com.google.android.gms.tasks.OnSuccessListener;
 
 import java.lang.ref.WeakReference;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
  * Sample LocationEngine using Google Play Services
  */
 class GoogleLocationEngine extends LocationEngine implements
-  GoogleApiClient.ConnectionCallbacks, GoogleApiClient.OnConnectionFailedListener, LocationListener {
+  GoogleApiClient.ConnectionCallbacks, GoogleApiClient.OnConnectionFailedListener {
 
   private WeakReference<Context> context;
   private GoogleApiClient googleApiClient;
+  private FusedLocationProviderClient fusedLocationProviderClient;
   private final Map<LocationEnginePriority, UpdateGoogleRequestPriority> REQUEST_PRIORITY = new
     HashMap<LocationEnginePriority, UpdateGoogleRequestPriority>() {
       {
@@ -62,6 +68,7 @@ class GoogleLocationEngine extends LocationEngine implements
       .addOnConnectionFailedListener(this)
       .addApi(LocationServices.API)
       .build();
+    fusedLocationProviderClient = LocationServices.getFusedLocationProviderClient(context);
   }
 
   static synchronized LocationEngine getLocationEngine(Context context) {
@@ -102,17 +109,28 @@ class GoogleLocationEngine extends LocationEngine implements
   public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
   }
 
+  @SuppressLint("MissingPermission")
   @Override
   public Location getLastLocation() {
     if (googleApiClient.isConnected()) {
+      final Location[] lastLocation = {null};
+
       //noinspection MissingPermission
-      return LocationServices.FusedLocationApi.getLastLocation(googleApiClient);
+      fusedLocationProviderClient.getLastLocation().addOnSuccessListener(new OnSuccessListener<Location>() {
+        @Override
+        public void onSuccess(Location location) {
+          lastLocation[0] = location;
+        }
+      });
+
+      return lastLocation[0];
     }
 
     //noinspection MissingPermission
     return null;
   }
 
+  @SuppressLint("MissingPermission")
   @Override
   public void requestLocationUpdates() {
     LocationRequest request = LocationRequest.create();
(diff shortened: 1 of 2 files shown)


liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -102,17 +109,28 @@ public void onConnectionSuspended(int cause) {
   public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
   }
 
+  @SuppressLint("MissingPermission")
```
tobrun: instead of suppressing Lint can we add the required permissions annotations instead?

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -102,17 +109,28 @@ public void onConnectionSuspended(int cause) {
...
       //noinspection MissingPermission
-      return LocationServices.FusedLocationApi.getLastLocation(googleApiClient);
+      fusedLocationProviderClient.getLastLocation().addOnSuccessListener(new OnSuccessListener<Location>() {
+        @Override
+        public void onSuccess(Location location) {
+          lastLocation[0] = location;
+        }
+      });
+
+      return lastLocation[0];
```
tobrun: method block above will always return null

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -102,17 +109,28 @@ public void onConnectionSuspended(int cause) {
...
     }
 
     //noinspection MissingPermission
     return null;
   }
 
+  @SuppressLint("MissingPermission")
```
tobrun: instead of suppressing Lint can we add the required permissions annotations instead?

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -8,22 +9,27 @@
...
 
 /**
  * Sample LocationEngine using Google Play Services
  */
 class GoogleLocationEngine extends LocationEngine implements
-  GoogleApiClient.ConnectionCallbacks, GoogleApiClient.OnConnectionFailedListener, LocationListener {
+  GoogleApiClient.ConnectionCallbacks, GoogleApiClient.OnConnectionFailedListener {
 
   private WeakReference<Context> context;
   private GoogleApiClient googleApiClient;
```
tobrun: GoogleApiClient is no longer required, all code related to it can be removed, a full blog post describing this https://android-developers.googleblog.com/2017/06/reduce-friction-with-new-location-apis.html

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -147,12 +165,19 @@ public Type obtainType() {
...
 
-  @Override
-  public void onLocationChanged(Location location) {
-    for (LocationEngineListener listener : locationListeners) {
-      listener.onLocationChanged(location);
+  private LocationCallback locationCallback = new LocationCallback() {
+    @Override
+    public void onLocationResult(LocationResult locationResult) {
+      List<Location> locationList = locationResult.getLocations();
+      if (locationList.size() > 0) {
```
tobrun: nit use `!list.isEmpty()` over `locationList.size() > 0`, depending on the List implementation this can result in  overhead.

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -102,17 +109,28 @@ public void onConnectionSuspended(int cause) {
   public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
   }
 
+  @SuppressLint("MissingPermission")
   @Override
   public Location getLastLocation() {
```
tobrun: I would suggest deprecating `getLastKnownLocation()` and  introducing a paradigm similar to Google by using success and failure listener so we have an api as `getLastKnowLocation(SomeSuccessCallback<Location>, SomeFailCallback<Error>)`. Introducing this change will also allow us to be more flexible in other implementations of LocationEngine.  if someone would require performing an async operation as part of this API, they would be able to vs null value returns.

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
```diff
@@ -100,6 +100,16 @@ public void onLocationChanged(Location location) {
     LocalBroadcastManager.getInstance(this).sendBroadcast(LocationReceiver.supplyIntent(location));
   }
 
+  @Override
+  public void onLastLocationSuccess(Location location) {
```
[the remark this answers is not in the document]
electrostat: These 2 are empty, because we are not using `LastLocation` within the telemetry service.

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -88,31 +97,39 @@ public boolean isConnected() {
...
-    }
-  }
-
-  @Override
-  public void onConnectionSuspended(int cause) {
-  }
-
-  @Override
-  public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
+  public void requestLastLocation() {
```
tobrun: Why not mimic the same callback system as google? Why would there be a need for multiple last location update listener? This callback would only be used once. No need to maintain a list, upstream API is able to handle multiple requests.

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -2,26 +2,34 @@
...
  * Sample LocationEngine using Google Play Services
  */
-class GoogleLocationEngine extends LocationEngine implements
-  GoogleApiClient.ConnectionCallbacks, GoogleApiClient.OnConnectionFailedListener, LocationListener {
+class GoogleLocationEngine extends LocationEngine implements GoogleApiClient.ConnectionCallbacks,
+  GoogleApiClient.OnConnectionFailedListener, LocationListener {
 
+  private FusedLocationProviderClient fusedLocationProviderClient;
   private WeakReference<Context> context;
   private GoogleApiClient googleApiClient;
```
tobrun: https://android-developers.googleblog.com/2017/06/reduce-friction-with-new-location-apis.html

shows that GoogleApiClient shouldn't be used anymore. Please remove

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -72,43 +70,43 @@ static synchronized LocationEngine getLocationEngine(Context context) {
...
 
+  @Deprecated
   @Override
   public Location getLastLocation() {
-    if (googleApiClient.isConnected()) {
-      //noinspection MissingPermission
-      return LocationServices.FusedLocationApi.getLastLocation(googleApiClient);
-    }
-
+    // No op
```
Guardiola31337: "Deprecating" doesn't mean to remove working functionality. What about using `fusedLocationProviderClient.getLastLocation().getResult()`? I believe that if connected will return the last location, `null` otherwise.

PR #172, opened 2018-07-03
Create Encrypted Database for Deferred Telemetry

electrostat:
Create an encrypted database for holding deferred telemetry data. Use a combination of `Rooms` and sql encryption.


alfwatt:
Stale

PR #173, opened 2018-07-05
Adjust Wakeup

electrostat:
Adjust system wakeup

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/SchedulerFlusherFactory.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/SchedulerFlusherFactory.java
@@ -3,16 +3,21 @@ package com.mapbox.android.telemetry;
 
 import android.app.AlarmManager;
 import android.content.Context;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager;
+import android.util.Log;
 
 class SchedulerFlusherFactory {
   static final String SCHEDULER_FLUSHER_INTENT = "com.mapbox.scheduler_flusher";
-  static final long FLUSHING_PERIOD_IN_MILLIS = 180 * 1000;
+  private static final String KEY_META_DATA_WAKE_UP = "com.mapbox.AdjustWakeUp";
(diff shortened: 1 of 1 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -548,4 +551,20 @@ void onEnterForeground() {
...
     ProcessLifecycleOwner.get().getLifecycle().removeObserver(this);
   }
+
+  private boolean checkWakeUp() {
+    try {
+      ApplicationInfo appInformation = applicationContext.getPackageManager()
+        .getApplicationInfo(applicationContext.getPackageName(),
+        PackageManager.GET_META_DATA);
+      if (appInformation != null && appInformation.metaData != null) {
+        boolean adjustWakeUp = appInformation.metaData.getBoolean(KEY_META_DATA_WAKE_UP, false);
```
LukasPaczos: I think the extraction of the `adjustWakeUp` flag could be moved to a method, not to duplicate the code in both classes. Does that make sense in this context?

libtelemetry/src/main/java/com/mapbox/android/telemetry/SchedulerFlusherFactory.java
```diff
@@ -25,4 +29,19 @@ SchedulerFlusher supply() {
...
   }
+
+  private void checkUpdatePeriod() {
+    try {
+      ApplicationInfo appInformation = context.getPackageManager().getApplicationInfo(context.getPackageName(),
+        PackageManager.GET_META_DATA);
+      if (appInformation != null && appInformation.metaData != null) {
+        boolean adjustWakeUp = appInformation.metaData.getBoolean(KEY_META_DATA_WAKE_UP, false);
+        if (adjustWakeUp) {
+          FLUSHING_PERIOD_IN_MILLIS = 600000;
```
LukasPaczos: Could you split this value into a more readable format? Also, the name of the value shouldn't be in all caps since it's mutable, for readability sake as well.

Issue #177, opened 2018-07-11
Change useragent in Maps SDK for Android traffic to `mapbox-maps-android`


lilykaiser:
Please change the useragent string for the Maps SDK for Android to `mapbox-maps-android` from `MapboxEventsAndroid`.

I did a quick spot check yesterday:
- iOS Maps SDK traffic has `mapbox-maps-ios` in the useragent
- Navigation uses `mapbox-navigation-android` or `mapbox-navigation-ios`

Changing the Android Maps SDK to `mapbox-maps-android` will bring it in line with our other mobile traffic.

Thanks!

zugaldia:
_If_ we decide to make this change, this string should be updated too: https://github.com/mapbox/mapbox-gl-native/blob/release-espresso/platform/android/MapboxGLAndroidSDK/build.gradle#L33

lilykaiser:
I discussed with @mick the week before last and confirmed that we would like to make this change to bring the Maps Android SDK in line with other user agents, and that making this change wouldn't break anything that exists in our back end today. I'll cut a ticket to update https://github.com/mapbox/mapbox-gl-native/blob/release-espresso/platform/android/MapboxGLAndroidSDK/build.gradle#L33

tobrun:
PR upstream -> https://github.com/mapbox/mapbox-gl-native/pull/12706, will ship with alpha v6.5.0 today

lilykaiser:
Thanks @tobrun ! Can https://github.com/mapbox/mapbox-events-android/pull/179 now be merged @zugaldia @electrostat ?

electrostat:
@lilykaiser should be able to be done, as long as I get a ✅

electrostat:
This was closed by: https://github.com/mapbox/mapbox-events-android/pull/179

PR #179, opened 2018-07-11
Add maps user agent

electrostat:
Add `mapbox-maps-android` user agent to list of acceptable user agents to standardize reporting across iOS and Android.

Fixes https://github.com/mapbox/mapbox-events-android/issues/177

cc: @lilykaiser


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -34,13 +34,15 @@
   private static final String UNITY_USER_AGENT = "MapboxEventsUnityAndroid/";
   private static final String NAVIGATION_USER_AGENT = "mapbox-navigation-android/";
   private static final String NAVIGATION_UI_USER_AGENT = "mapbox-navigation-ui-android/";
+  private static final String MAPS_USER_AGENT = "mapbox-maps-android";
```
tobrun: one comment though is that other user-agent string end with `/` while the maps doesn't

electrostat: Good catch

PR #184, opened 2018-07-18
Crash When Disabling Location Permissions Manually

electrostat:
Our SDK currently crashes when a user disables location permissions manually (through the Android settings) and returns to the parent application. LocationEngine is attempting to acquire location without permission. 

Change was to add in both pre and post api 23 permission check before activating the engine. 

Fixes https://github.com/mapbox/mapbox-events-android/issues/182#issuecomment-405951475

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
@@ -1,19 +1,24 @@
 package com.mapbox.android.telemetry;
 
 
+import android.Manifest;
 import android.app.Service;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.content.pm.PackageManager;
 import android.location.Location;
 import android.os.Binder;
+import android.os.Build;
 import android.os.IBinder;
 import android.support.annotation.Nullable;
 import android.support.v4.content.LocalBroadcastManager;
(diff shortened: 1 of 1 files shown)


ufocom:
Hi, where/when i can find fix for this little bug?

electrostat:
@ufocom I just pushed this. You can check the latest  4.0.0-Snapshot of the Telemetry library [LINK](https://oss.sonatype.org/content/repositories/snapshots/com/mapbox/mapboxsdk/mapbox-android-telemetry/). We will work on getting it into the next maps sdk release.

ufocom:
@electrostat  so if i will add telemetry lib to my gradly problem will be solved?

electrostat:
@ufocom yes. As long as you exclude the telemetry and core libraries from the maps sdk. I will be doing a release today and then the maps team is looking to incorporate that into a release they are going to do shortly. So you can either exclude and add or wait for the next release.

ursimon:
to be more specific, this has resolved my issue:
```    
implementation('com.mapbox.mapboxsdk:mapbox-android-sdk:6.1.3') {
        exclude group: 'com.mapbox.android.core'
        exclude group: 'com.mapbox.android.telemetry'
    }
    implementation 'com.mapbox.mapboxsdk:mapbox-android-telemetry:3.2.0-SNAPSHOT'
```

Issue #181, opened 2018-07-14
java.lang.NullPointerException in com.mapbox.android.telemetry.TelemetryUtils.obtainSharedPreferences on Android 8.1


Crashthatch:
Seeing this crash in Google Play Console crash reports since I upgraded to the latest version.
8 crash reports so far, all on Android 8.1, using mapbox-events-android:3.1.4, being called from within Unity.

```
java.lang.RuntimeException: 
  at android.app.ActivityThread.handleStopService (ActivityThread.java:3666)
  at android.app.ActivityThread.-wrap26 (Unknown Source)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1796)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:164)
  at android.app.ActivityThread.main (ActivityThread.java:6753)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:482)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:807)
Caused by: java.lang.NullPointerException: 
  at com.mapbox.android.telemetry.TelemetryUtils.obtainSharedPreferences (TelemetryUtils.java:191)
  at com.mapbox.android.telemetry.TelemetryLocationEnabler.updateLocationPreferences (TelemetryLocationEnabler.java:61)
  at com.mapbox.android.telemetry.TelemetryLocationEnabler.updateTelemetryLocationState (TelemetryLocationEnabler.java:40)
  at com.mapbox.android.telemetry.TelemetryService.disableTelemetryLocationState (TelemetryService.java:243)
  at com.mapbox.android.telemetry.TelemetryService.onDestroy (TelemetryService.java:65)
  at android.app.ActivityThread.handleStopService (ActivityThread.java:3648)
```

electrostat:
I am currently investigating this and will get back to you, when I have more information to share.

electrostat:
@Crashthatch I've am addressing this within this PR: https://github.com/mapbox/mapbox-events-android/pull/190 . Looks to be a result of a null context under certain circumstances. I've added a check to the system to prevent this and get an appropriate context.

Crashthatch:
Looks like your fix should work to me. I'll update after the next release and keep an eye on the live crash logs for a few days to check this is resolved.

Issue #146, opened 2018-05-16
Root Cause of Missing Context


electrostat:
Investigate missing `MapboxTelemetry.applicationContext` when generating a location event.

- possible causes could be a race condition or odd background state
- solution may be to just drop events when context is not set

Reference Ticket: https://github.com/mapbox/mapbox-events-android/pull/144

cc: @zugaldia

electrostat:
After researching and investigating this situation, there is a big hole within our Telemetry sdk that could cause this crash. The hole results from event creation and the Telemetry Service from being decoupled from the `MapboxTelemetry` class. This means we can generate events (before pushing them) without `applicationContext` being set and `TelemetryService` could be generated and function without the context set. 

We already have a fix in that piggy-backs off of the ability for a service to get `applicationContext` to set `MapboxTelemetry.applicationContext` when it is null. So that is solved for us.

We need a solution that makes sure `applicationContext` is set when generating events. This is twofold here. This problem will take care of itself when we pull all of the event generation code out of our sdk and put it into the parent sdks for handling. There are some events (like AppUserTurnstile) and utilities (nav and telem) which will likely not be moving over. We will need a solution for those and a possible short-term solution for the other events. 

My recommendation is we make those utils and event creation dependent on `MapboxTelemetry`. A parent sdk/app could only receive the utils and event factories through a fully initialized MapboxTelemetry class.

electrostat:
Researched #3 from above and all options require context to be used.

electrostat:
I pulled in all the relevant tickets from previous instances of this context being null and causing issues. Putting them all here for easy access and review when solution has been made.

https://github.com/mapbox/mapbox-gl-native/issues/12748
https://github.com/mapbox/mapbox-events-android/pull/144
https://github.com/mapbox/mapbox-events-android/issues/201 -- will include in PR
https://github.com/mapbox/mapbox-events-android/issues/188
https://github.com/mapbox/mapbox-events-android/issues/181
https://github.com/mapbox/mapbox-events-android/pull/190
https://github.com/mapbox/mapbox-events-android/issues/170

zugaldia:
Can we start with adding null check guards while we work on a longer term solution? I'd be interested in a fix able to make it to the upcoming Maps `f` release (due Sep 12).

electrostat:
[...]

PR #190, opened 2018-07-24
checkApplicationContext when stopping service

electrostat:
Add `checkApplicationContext()` to workflow of stopping `TelemetryService`. This will prevent a NPE when attempting to run processes on service shutdown under certain circumstances.

Fixes https://github.com/mapbox/mapbox-events-android/issues/181

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
@@ -66,6 +66,7 @@ public class TelemetryService extends Service implements TelemetryCallback, Loca
 
   @Override
   public void onDestroy() {
+    checkApplicationContext();
     unregisterLocationReceiver();
     unregisterTelemetryReceiver();
     disableTelemetryLocationState();


danesfeder:
Also noting that we are using a combination of `MapboxTelemetry.applicationContext` and `getApplication()` in `TelemetryService`.  It may make more sense to align which one is used to avoid confusion around NPEs like the one this PR resolves.

Guardiola31337:
The problem here is https://github.com/mapbox/mapbox-events-android/blob/a57bf695443c9c20bc593e92af0c7430078384f5/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java#L64
`applicationContext` is `static` and suddenly (under some scenarios / specific devices) gets nullified - causing this kind of crashes when reaching methods in which the context is used. So in order to avoid them, we need to ensure that the `applicationContext` is not `null` at those points.

But yeah, I agree that we should 👀 into this thoroughly and try to be more consistent and less defensive across the codebase.

cc @electrostat @danesfeder @LukasPaczos

electrostat:
@Guardiola31337 looking into this within this ticket: https://github.com/mapbox/mapbox-events-android/issues/146

Issue #159, opened 2018-06-05
Use IntentService with PendingIntent to receive background location updates


tobrun:
Capturing from [LocationRequest documentation](https://developers.google.com/android/reference/com/google/android/gms/location/FusedLocationProviderClient#requestLocationUpdates(com.google.android.gms.location.LocationRequest,%20android.app.PendingIntent)):

> This method is suited for the background use cases, more specifically for receiving location updates, even when the app has been killed by the system. In order to do so, use a PendingIntent for a started service. For foreground use cases, the LocationCallback version of the method is recommended, see requestLocationUpdates(LocationRequest, LocationCallback, Looper).

We currently rely on using the other method above which should be used for foreground use-cases.

tobrun:
cc @mapbox/mobile-telemetry

andrlee:
@tobrun now that liblocation supports both batching and background mode. can this ticket be closed? it is not super clear to me whether the scope of it also covers integration into libtelemetry? thoughts?

andrlee:
Fixed by https://github.com/mapbox/mapbox-events-android/pull/236

Issue #163, opened 2018-06-14
Batch location events


tobrun:
To optimize background location update. It's advised to use batching of location updates.

> If your app is running in the background, the location system service computes a new location for your app only a few times each hour. This is the case even when your app is requesting more frequent location updates. By using the batched version of FLP, however, you have access to more time-frequent location history after your app receives a batch update

See related javadoc [here](https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest#setMaxWaitTime(long)).

Refs #159 

cc @electrostat @zugaldia

andrlee:
Fixed by https://github.com/mapbox/mapbox-events-android/pull/236

PR #192, opened 2018-07-27
Switch to PendingIntent System

electrostat:
Change from using Callback to utilizing PendingIntent for location updates with GoogleLocationEngine

Fixes https://github.com/mapbox/mapbox-events-android/issues/159#issuecomment-394836236

### DIFF
--- a/liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProvider.java
+++ b/liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProvider.java
@@ -82,7 +82,8 @@ public class LocationEngineProvider {
     backgroundLocationEngineDictionary = new HashMap<>();
     Map<LocationEngine.Type, LocationEngineSupplier> backgroundLocationEnginesDictionary =
       obtainDefaultBackgroundLocationEnginesDictionary();
-    for (Map.Entry<LocationEngine.Type, LocationEngineSupplier> entry : backgroundLocationEnginesDictionary.entrySet()) {
+    for (Map.Entry<LocationEngine.Type,
+      LocationEngineSupplier> entry : backgroundLocationEnginesDictionary.entrySet()) {
       LocationEngineSupplier locationEngineSupplier = entry.getValue();
       if (locationEngineSupplier.hasDependencyOnClasspath()) {
         LocationEngine available = locationEngineSupplier.supply(context);


liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -170,4 +173,26 @@ private void connect() {
   private void updateRequestPriority(LocationRequest request) {
     REQUEST_PRIORITY.get(priority).update(request);
   }
+
+  private PendingIntent getPendingIntent() {
+    if (context.get() != null) {
+
+      if (Build.VERSION.SDK_INT >= 26) {
```
danesfeder: I think we can clean up this logic by adding something like a `LocationIntentProvider`.  This class could be in charge of providing the `PendingIntent` based on the `Build.VERSION.SDK_INT` so this code would look like:

```
Context context = context.get();
if (context != null) {
    SdkChecker checker = new SdkChecker();
    LocationIntentProvider provider = new LocationIntentProvider(context, sdkChecker, locationListeners);
    Intent intent = provider.buildIntent();
    return PendingIntent.getService(context.get(), 0, intent, PendingIntent.FLAG_UPDATE_CURRENT);
}
```

`SdkChecker` would be a simple class for checking `Build.VERSION.SDK_INT >= 26` with a function like `SdkChecker#isOreoOrAbove()` 

The reason behind all this is for testing - so we can "mock" the build version and see which `PendingIntent` the provider returns. 

```
@Test
public void whenApiVersionIsOreo_providerReturnsBroadcastReceiver() {
    SdkChecker checker = mock(SdkChecker.class);
    when(checker.isOreoOrAbove()).thenReturn(true);
    ...
}
```

liblocation/src/main/java/com/mapbox/android/core/location/LocationBroadcastReceiver.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class LocationBroadcastReceiver extends BroadcastReceiver {
+
+  static final String ACTION_PROCESS_UPDATES =
+    "com.mapbox.android.core.location.LocationBroadcastReceiver.ACTION_PROCESS_UPDATES";
+  private static CopyOnWriteArrayList<LocationEngineListener> locationEngineListeners;
+
+  @Override
+  public void onReceive(Context context, Intent intent) {
+    if (intent != null) {
```
danesfeder: This block of logic is duplicated in the `IntentService` as well, maybe we can extract into a class like `LocationIntentHandler` that takes the listeners as a param?  This can be passed to be used is the service or receiver.

```
LocationIntentHandler intentHandler = new LocationIntentHandler(locationEngineListeners);
LocationUpdateBroadcastReceiver.addIntentHandler(intentHandler);
```

Then `onReceive` or `onHandleIntent` would end up looking like:
```
@Override
public void onReceive(Context context, Intent intent) {
    intentHandler.handle(intent);
}
```

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -170,4 +173,26 @@ private void connect() {
...
+      } else {
+        Intent intent = new Intent(context.get(), FusedLocationIntentService.class);
+        intent.setAction(FusedLocationIntentService.ACTION_PROCESS_UPDATES);
+
+        FusedLocationIntentService.addListeners(locationListeners);
+
+        return PendingIntent.getService(context.get(), 0, intent, PendingIntent.FLAG_UPDATE_CURRENT);
+      }
+    }
+    return null;
```
danesfeder: Are we okay to return `null` here?

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -170,4 +172,16 @@ private void connect() {
...
+
+  private PendingIntent getPendingIntent() {
+    Context context = this.context.get();
+    if (context != null) {
+      SdkChecker sdkChecker = new SdkChecker();
+      LocationPendingIntentProvider provider =
+        new LocationPendingIntentProvider(context, sdkChecker, locationListeners);
+      return provider.buildPendingIntent();
+    } else {
+      throw new IllegalStateException(CONTEXT_NULL);
```
danesfeder: On second thought here we can build the `PendingIntent` in the constructor I believe, so we don't have to worry about the `WeakReference` being `null`

liblocation/src/main/java/com/mapbox/android/core/location/LocationPendingIntentProvider.java
```diff
@@ -0,0 +1,40 @@
...
+  private CopyOnWriteArrayList<LocationEngineListener> locationEngineListeners;
+
+  LocationPendingIntentProvider(Context context, SdkChecker sdkChecker,
+                         CopyOnWriteArrayList<LocationEngineListener> locationEngineListeners) {
+    this.context = context;
+    this.sdkChecker = sdkChecker;
+    this.locationEngineListeners = locationEngineListeners;
+  }
+
+  PendingIntent buildPendingIntent() {
```
danesfeder: We can build the `PendingIntent` in the constructor so we don't have to store the `Context`, checker, etc. in this class.

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -134,14 +133,14 @@ public void requestLocationUpdates() {
 
     if (googleApiClient.isConnected()) {
```
florianPOLARSTEPS: Hey @electrostat .
Great work on this one! Not sure if this pull request should also already fix https://github.com/mapbox/mapbox-events-android/issues/163 as well, but unless you set LocationRequest.setMaxWaitTime(long) you will not receive location updates in a batched mode. (https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest.html#setMaxWaitTime(long))

Cheers

danesfeder: Hey @florianPOLARSTEPS 👋 this PR is aiming to unlock the work for #163, where we will then complete the `Location` batching setup.  Thanks for the quick feedback here and stay tuned as we get work going on that PR.

danesfeder:
@Guardiola31337 do you mind double checking how we are passing the `LocationEngineListener` list here?  I want to make sure that we are ensuring the list is correct when the `LocationIntentHandler` receives new `Intent`s

liblocation/src/main/java/com/mapbox/android/core/location/SdkChecker.java
```diff
@@ -0,0 +1,10 @@
+package com.mapbox.android.core.location;
+
+import android.os.Build;
+
+public class SdkChecker {
+
+  boolean isOreoOrAbove() {
+    return Build.VERSION.SDK_INT >= 26;
```
LukasPaczos: You can use a constant instead of a raw number here.

liblocation/src/main/java/com/mapbox/android/core/location/SdkChecker.java
```diff
@@ -0,0 +1,10 @@
+package com.mapbox.android.core.location;
+
+import android.os.Build;
+
+public class SdkChecker {
+
+  boolean isOreoOrAbove() {
```
LukasPaczos: This method can be static.

liblocation/src/main/java/com/mapbox/android/core/location/LocationPendingIntentProvider.java
```diff
@@ -0,0 +1,49 @@
...
+  LocationPendingIntentProvider(Context context, SdkChecker sdkChecker,
+                                CopyOnWriteArrayList<LocationEngineListener> locationEngineListeners) {
+    locationPendingIntent = buildIntent(context, sdkChecker, locationEngineListeners);
+  }
+
+  LocationPendingIntent intent() {
+    return locationPendingIntent;
+  }
+
+  private LocationPendingIntent buildIntent(Context context, SdkChecker sdkChecker,
```
LukasPaczos: It feels like this method can also be static, I can't see we would want to hold a reference to the `LocationPendingIntentProvider`.

electrostat:
@LukasPaczos since we are passing off the service and broadcast receiver to the FusedLocationApi to manage and control, these 3 points don't seem possible. I tested and attempted to force the creation of separate instances of the receiver and service to see if they would be triggered by the system. It's all in the OS's hands at this point. Same issue with moving the service to a different thread.

If you see something else, let me know.

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -1,31 +1,35 @@
...
 import com.google.android.gms.location.LocationRequest;
 import com.google.android.gms.location.LocationServices;
 
-import java.lang.ref.WeakReference;
 import java.util.HashMap;
 import java.util.Map;
 
 /**
- * Sample LocationEngine using Google Play Services
+ * Sample LocationEngine using Google Play Servicesre
```
LukasPaczos: Typo

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -170,4 +169,22 @@ private void connect() {
...
-}
+
+  private void generatePendingIntent(Context context) {
+    pendingIntent = LocationPendingIntentProvider.buildIntent(context, this).retrievePendingIntent();
+  }
+
+  private void setupBroadcastReceiver(Context context) {
+    final String action = "GoogleLocationEngineBroadcast-" + this.toString();
+    IntentFilter filter = new IntentFilter();
+    filter.addAction(action);
```
LukasPaczos: NIT: this can be passed in the constructor.

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -170,4 +169,22 @@ private void connect() {
...
     REQUEST_PRIORITY.get(priority).update(request);
   }
-}
+
+  private void generatePendingIntent(Context context) {
+    pendingIntent = LocationPendingIntentProvider.buildIntent(context, this).retrievePendingIntent();
+  }
+
+  private void setupBroadcastReceiver(Context context) {
+    final String action = "GoogleLocationEngineBroadcast-" + this.toString();
```
LukasPaczos: I think using just `hashCode` is safer, in case `toString` is overwritten at any time in the future. This value can also be stored as the field and passed to the intent provider.

liblocation/src/main/java/com/mapbox/android/core/location/LocationPendingIntentProvider.java
```diff
@@ -0,0 +1,32 @@
...
+import android.content.Context;
+import android.content.Intent;
+
+import java.util.Random;
+
+class LocationPendingIntentProvider {
+
+  private final LocationPendingIntent locationPendingIntent;
+
+  LocationPendingIntentProvider(Context context, GoogleLocationEngine googleLocationEngine) {
```
LukasPaczos: This seems to be unused.

liblocation/src/main/java/com/mapbox/android/core/location/LocationPendingIntentProvider.java
```diff
@@ -0,0 +1,32 @@
...
+    locationPendingIntent = buildIntent(context, googleLocationEngine);
+  }
+
+  LocationPendingIntent intent() {
```
LukasPaczos: This seems to be unused.

liblocation/src/main/java/com/mapbox/android/core/location/LocationPendingIntentProvider.java
```diff
@@ -0,0 +1,32 @@
...
+  static LocationPendingIntent buildIntent(Context context, GoogleLocationEngine googleLocationEngine) {
+    Intent intent = new Intent("GoogleLocationEngineBroadcast-" + googleLocationEngine.toString());
+
+    PendingIntent broadcastPendingIntent = PendingIntent.getBroadcast(context, generateRequestCode(), intent, 0);
+    return new LocationBroadcastPendingIntent(broadcastPendingIntent);
+  }
+
+  private static int generateRequestCode() {
+    Random random = new Random();
+    return random.nextInt();
```
LukasPaczos: We need to statically keep track of all of the registered request codes, otherwise, there's a chance that we will overwrite one of the requests when creating a new instance of the engine.

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -170,4 +169,22 @@ private void connect() {
...
+    IntentFilter filter = new IntentFilter();
+    filter.addAction(action);
+    this.broadcastReceiver = new BroadcastReceiver() {
+      @Override
+      public void onReceive(Context context, Intent intent) {
+        locationIntentHandler.handle(intent, action);
+      }
+    };
+
+    context.registerReceiver(broadcastReceiver, filter);
```
LukasPaczos: This broadcast receiver should be created in the constructor and registered/unregistered whenever location updates are requested/removed.

liblocation/src/main/java/com/mapbox/android/core/location/LocationPendingIntentProvider.java
```diff
@@ -0,0 +1,32 @@
...
+  LocationPendingIntentProvider(Context context, GoogleLocationEngine googleLocationEngine) {
+    locationPendingIntent = buildIntent(context, googleLocationEngine);
+  }
+
+  LocationPendingIntent intent() {
+    return locationPendingIntent;
+  }
+
+  static LocationPendingIntent buildIntent(Context context, GoogleLocationEngine googleLocationEngine) {
+    Intent intent = new Intent("GoogleLocationEngineBroadcast-" + googleLocationEngine.toString());
```
LukasPaczos: We can pass the `action` string instead of building it from scratch with the engine's instance.

liblocation/src/main/java/com/mapbox/android/core/location/LocationIntentHandler.java
```diff
@@ -0,0 +1,34 @@
...
+  private CopyOnWriteArrayList<LocationEngineListener> locationListeners;
+
+  LocationIntentHandler(CopyOnWriteArrayList<LocationEngineListener> locationListeners) {
+    this.locationListeners = locationListeners;
+  }
+
+  void handle(Intent intent, String actionTag) {
+    if (intent != null) {
+      final String action = intent.getAction();
+      if (actionTag.equals(action)) {
```
LukasPaczos: I believe we don't need to compare action tags here, this should be handled by the `IntentFilter`.

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngine.java
```diff
@@ -133,15 +138,16 @@ public void requestLocationUpdates() {
     updateRequestPriority(request);
 
     if (googleApiClient.isConnected()) {
+      locationIntentHandler = new LocationIntentHandler(locationListeners);
```
LukasPaczos: We don't need a new instance of the handler whenever location updates are requested. I think this should be done once in the constructor.

liblocation/src/main/java/com/mapbox/android/core/location/LocationPendingIntent.java
```diff
@@ -0,0 +1,8 @@
+package com.mapbox.android.core.location;
+
+import android.app.PendingIntent;
+
+interface LocationPendingIntent {
+
+  PendingIntent retrievePendingIntent();
```
LukasPaczos: I'm not seeing the `IntentService` setup anymore, do we still need this interface?

electrostat: We are still using it for the BroadcastReceiver

liblocation/src/main/java/com/mapbox/android/core/location/LocationPendingIntentProvider.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.ArrayList;
+import java.util.Random;
+
+class LocationPendingIntentProvider {
+  private static ArrayList<Integer> requestCodes = new ArrayList<>();
+
+  static LocationPendingIntent buildIntent(Context context, String action) {
+    Intent intent = new Intent(action);
+
+    PendingIntent broadcastPendingIntent = PendingIntent.getBroadcast(context, generateRequestCode(), intent, 0);
```
LukasPaczos: The `requestCodes` list is never updated with the generated code so the `duplicateCode` check is not behaving as expected. Also, to make this solution clean, the `requestCodes` list should be updated whenever we request/remove location updates.

electrostat:
@tobrun this works in the foreground. I've tested it out and am getting location updates just fine. I would've brought up this as a concern earlier, if it  did not work that way. Also, the goal of this system is to prevent any issues for the host app regarding location engines in general, as well as our implementation of that engine within the telemetry sdk.

If you aren't getting location data when testing this, then I can dig in and see why that's happening.

tobrun:
@electrostat it's not about if this works or not. It's about following the guidelines and they show you to make a distinction between foreground and background use-cases.

> requestLocationUpdates (LocationRequest request, LocationCallback callback, Looper looper)
Requests location updates with a callback on the specified Looper thread. This method is suited for the foreground use cases. For background use cases, the PendingIntent version of the method is recommended, see requestLocationUpdates(LocationRequest, PendingIntent).

PR #193, opened 2018-08-03
Offline maps event

zugaldia:
This PR adds a new `MapOfflineEvent` event to the Maps SDK list of events.

It's still a WIP - not all fields are finalized, and tests are missing.

cc: @electrostat @Guardiola31337 for early eyes.

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapEventFactory.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapEventFactory.java
@@ -21,6 +21,7 @@ public class MapEventFactory {
   private static final int NO_NETWORK = -1;
   private static final String NOT_A_LOAD_MAP_EVENT_TYPE = "Type must be a load map event.";
   private static final String NOT_A_GESTURE_MAP_EVENT_TYPE = "Type must be a gesture map event.";
+  private static final String NOT_AN_OFFLINE_MAP_EVENT_TYPE = "Type must be an offline map event.";
   private static final String MAP_STATE_ILLEGAL_NULL = "MapState cannot be null.";
   private static final Map<Integer, String> ORIENTATIONS = new HashMap<Integer, String>() {
     {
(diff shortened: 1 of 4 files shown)


asheemmamoowala:
@zugaldia  [#11447](https://github.com/mapbox/mapbox-gl-native/pull/11447) adds support for shapes to define offline regions. In addition to the bounding box, it would be helpful if the telemetry data also collects the zoom range and geometry type (for shape-based regions).  

With the geometry type, it would be helpful to understand how developers use the offline regions  - for example - routes vs admin boundaries.

zugaldia:
@asheemmamoowala totally, thanks for linking the ongoing work. For now this WIP PR is targeting the fields in [`OfflineRegionDefinition`](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionDefinition.java) which is the base interface we use for all regions, and it only has knowledge of bounds. I'm definitely interested in adding more fields, to support https://github.com/mapbox/mapbox-gl-native/pull/11447 and the current `OfflineTilePyramidRegionDefinition`.

osana:
Related PR in event-schema repo : https://github.com/mapbox/event-schema/pull/64

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapEventFactory.java
```diff
@@ -21,6 +21,7 @@
   private static final int NO_NETWORK = -1;
   private static final String NOT_A_LOAD_MAP_EVENT_TYPE = "Type must be a load map event.";
   private static final String NOT_A_GESTURE_MAP_EVENT_TYPE = "Type must be a gesture map event.";
+  private static final String NOT_AN_OFFLINE_MAP_EVENT_TYPE = "Type must be an offline map event.";
```
electrostat: Can be removed, since we are no longer worried about type for this event during creation.

PR #202, opened 2018-08-28
Add null checks for obtainSharedPreferences

electrostat:
Add null checks within parent methods for `obtainSharedPreferences` 

Fixes https://github.com/mapbox/mapbox-gl-native/issues/12748

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
@@ -177,14 +177,18 @@ public class TelemetryUtils {
   }
 
   static String retrieveVendorId() {
-    SharedPreferences sharedPreferences = obtainSharedPreferences();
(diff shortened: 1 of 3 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
```diff
@@ -177,6 +177,10 @@ static boolean isEmpty(String string) {
   }
 
   static String retrieveVendorId() {
+    if (MapboxTelemetry.applicationContext == null) {
+      return updateVendorId();
```
Guardiola31337: Have we considered returning `obtainUniversalUniqueIdentifier()` here? Doing so, would remove the need for https://github.com/mapbox/mapbox-events-android/blob/329a1d657bbcf6ff555d64ef028d1a16f2c1d3ec/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java#L199-L203 in `updateVendorId()` 👇

Issue #207, opened 2018-09-10
java.lang.ClassCastException: android.os.BinderProxy cannot be cast to com.mapbox.android.telemetry.TelemetryService$TelemetryBinder


andrlee:
Dups: https://github.com/mapbox/mapbox-gl-native/issues/12460

PR #209, opened 2018-09-10
Detect BinderProxy

electrostat:
Need to detect when the system sends a `BinderProxy` back instead of our actual service binder, once services is connected.

Fixes https://github.com/mapbox/mapbox-events-android/issues/207


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -390,14 +390,18 @@ private ServiceConnection obtainServiceConnection() {
     return new ServiceConnection() {
       @Override
       public void onServiceConnected(ComponentName className, IBinder service) {
-        TelemetryService.TelemetryBinder binder = (TelemetryService.TelemetryBinder) service;
-        telemetryService = binder.obtainService();
-        telemetryService.addServiceTask(MapboxTelemetry.this);
-        if (telemetryService.obtainBoundInstances() == 0) {
-          telemetryService.injectEventsQueue(queue);
+        if (service.getClass() == TelemetryService.TelemetryBinder.class) {
```
andrlee: Would these be equivalent to what you're doing on line 393: 
option A: `if (service instanceof TelemetryService.TelemetryBinder)`
option B: `TelemetryService.TelemetryBinder.class.isAssignableFrom(service.getClass())`
What are pros and cons of these approaches / performance implications?

Issue #210, opened 2018-09-11
Obtain best location engine always returns `GoogleLocationEngine`


msh91:
We have an issue with `LocationEngineProvider`. It always returns `GoogleLocationEngine` as the best engine, but in some situations like when 'google play services' is not available or is out of date, so we can use `AndroidLocationEngine` instead.

msh91:
I've created a [PR](https://github.com/mapbox/mapbox-events-android/pull/211) to add this feature.
Please let me know your opinion about it.

PR #211, opened 2018-09-11
Check google play services availability for providing best location engine #210

msh91:
When 'google play services' is not available or is out of date, we can use AndroidLocationEngine instead.
Issue #210


liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProvider.java
```diff
@@ -67,13 +75,35 @@ private void initAvailableLocationEngines(Context context) {
...
+    // Check to see if google play services is installed and updated to required version
+    if (googlePlayAvailabilityResultCode == ConnectionResult.SUCCESS) {
+      bestLocationEngine = getGoogleLocationEngine();
+    } else {
+      bestLocationEngine = getAndroidLocationEngine();
+    }
+    return bestLocationEngine;
+  }
+
+  @Nullable
```
yunikkk: Is it really `Nullable`? `obtainBestLocationEngineAvailable ` which calls `obtainBestLocationEngine` then returns `NonNull`

msh91: It's because of `obtainLocationEngineBy(type)` which is `Nullable`. I think method `getGoogleLocationEngine` should be `Nullable` too, I forgot to annotate it.
But if we are sure that `obtainLocationEngineBy` is NonNull then we can remove `Nullable` annotations.

yunikkk: I'd suggest that `LocationEngine.Type.ANDROID` is always present while `LocationEngine.Type.GOOGLE_PLAY_SERVICES` is not since it depends on google services status.

msh91:
@yunikkk Thanks for comment, I removed it.

liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProvider.java
msh91: Thanks for the comment, I removed it.

msh91:
any news about this pull request ?!

yunikkk:
> any news about this pull request ?!

Maybe @electrostat can help here?

msh91:
@electrostat No urgency from me too. We can wait for the new release. 
Thank you.

PR #215, opened 2018-09-12
Update MOCK LocationEngine Type to REPLAY

danesfeder:
With https://github.com/mapbox/mapbox-navigation-android/pull/1039, the navigation repository is removing the `MockLocationEngine` in favor of a `ReplayRouteLocationEngine`.  

This PR is just updating the `Type` `enum` to better reflect the new naming on our side.  If this doesn't make sense or there is a better way to address this, I'm happy to update this PR or close.  

cc @andrlee

### DIFF
--- a/liblocation/src/main/java/com/mapbox/android/core/location/LocationEngine.java
+++ b/liblocation/src/main/java/com/mapbox/android/core/location/LocationEngine.java
@@ -16,7 +16,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
  */
 public abstract class LocationEngine {
   public enum Type {
-    GOOGLE_PLAY_SERVICES, ANDROID, MOCK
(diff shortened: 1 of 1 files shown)


Guardiola31337:
Noting that this PR breaks SemVer and would require a major version bump.

IMO as commented in https://github.com/mapbox/mapbox-navigation-android/pull/1039#discussion_r216960076
> BTW at the end of the day this is a "mock" location engine anyways, right? For now, I think is fine, we can cut a ticket later upstream if needed.

the locations are ultimately "fake" so the naming here is not a problem, the "general" meaning of the type is untouched i.e. `MOCK` could mean "replay" too.

I'd suggest adding a new type `REPLAY` (instead of replacing the `MOCK` one as this PR did) if you think this is really necessary or rolling back this PR and leave it as was.

cc @electrostat @danesfeder @andrlee

PR #216, opened 2018-09-13
Re-add MOCK LocationEngine Type

danesfeder:
Per @Guardiola31337's comment > https://github.com/mapbox/mapbox-events-android/pull/215#issuecomment-420947181 it might make sense to re-add the `MOCK` type to avoid a SEMVER major breakage.  

Ultimately up to ya'll @electrostat @andrlee, thanks for the 👀

### DIFF
--- a/liblocation/src/main/java/com/mapbox/android/core/location/LocationEngine.java
+++ b/liblocation/src/main/java/com/mapbox/android/core/location/LocationEngine.java
@@ -16,7 +16,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
  */
 public abstract class LocationEngine {
   public enum Type {
-    GOOGLE_PLAY_SERVICES, ANDROID, REPLAY
+    GOOGLE_PLAY_SERVICES, ANDROID, MOCK
   }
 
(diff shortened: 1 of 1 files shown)


electrostat:
Yeah. It makes sense to not break `SemVer` until we have bigger changes we want to make and can remove this along with those.

liblocation/src/main/java/com/mapbox/android/core/location/LocationEngine.java
```diff
@@ -16,7 +16,7 @@
  */
 public abstract class LocationEngine {
   public enum Type {
-    GOOGLE_PLAY_SERVICES, ANDROID, REPLAY
+    GOOGLE_PLAY_SERVICES, ANDROID, MOCK, REPLAY
```
electrostat: I think we should mark `MOCK` as Deprecated. That way we can internally mark it for removal at the next major release and flag to others that it's going away.

andrlee: What's the benefit of adding a redundant enum? If we have a choice i'd not encourage changes on api signatures that are unnecessary.

Issue #199, opened 2018-08-22
Too generic proguard keep rule


linnar:
Proguard rule changed in #189 keeps almost everything from Google Play Services libraries resulting in around 6 thousand additional methods kept in our apk. Is it possible to replace this rule with more targeted ones?

PR #81, opened 2018-03-09
Update Proguard rules

electrostat:
Add in rules to handle warnings when utilizing proguard (unable to find certain play service classes) in builds

PR #217, opened 2018-09-19
More specific Proguard rules

electrostat:
Update proguard rules to be more specific and lower method counts. We utilize `common` and `location` within the `Core` library and will keep only those.

fixes https://github.com/mapbox/mapbox-events-android/issues/199

### DIFF
--- a/libcore/proguard-consumer.pro
+++ b/libcore/proguard-consumer.pro
@@ -1,6 +1,4 @@
 # Consumer proguard rules for libcore
 
 # --- GMS ---
--keep class com.google.android.gms.location.** { *; }
--keep class com.google.android.gms.common.** { *; }
 -dontwarn com.google.android.gms.**
--- a/libcore/proguard-consumer.pro
+++ b/libcore/proguard-consumer.pro
@@ -1,5 +1,6 @@
 # Consumer proguard rules for libcore
 
 # --- GMS ---
--keep class com.google.android.gms.** { *; }
+-keep class com.google.android.gms.location.** { *; }
+-keep class com.google.android.gms.common.** { *; }
 -dontwarn com.google.android.gms.**


libcore/proguard-consumer.pro
```diff
@@ -1,5 +1,6 @@
 # Consumer proguard rules for libcore
 
 # --- GMS ---
--keep class com.google.android.gms.** { *; }
+-keep class com.google.android.gms.location.** { *; }
+-keep class com.google.android.gms.common.** { *; }
```
tobrun: what is the reason for keeping these classes? Are you using reflection on these or any other reason why they should be kept?

electrostat: Per @LukasPaczos comment on the original PR that started this `Proguard` change (https://github.com/mapbox/mapbox-events-android/pull/81#discussion_r174079424). This is necessary to prevent warnings in the maps sdk when building for release.

LukasPaczos: True. However, we should only keep the required ones, or maybe we can even `dontwarn` some of those?

electrostat: We could remove the `keep` and then keep the `dontwarn`. Does that implementation work for you two, without having any negative effects on the maps sdk?

LukasPaczos: Refs https://developers.google.com/android/guides/setup:
> Note: ProGuard directives are included in the Play services client libraries to preserve the required classes. The Android Plugin for Gradle automatically appends ProGuard configuration files in an AAR (Android ARchive) package and appends that package to your ProGuard configuration. During project creation, Android Studio automatically creates the ProGuard configuration files and build.gradle properties for ProGuard use. To use ProGuard with Android Studio, you must enable the ProGuard setting in your build.gradle buildTypes. For more information, see the ProGuard guide.

What I understand from there is that if you are not referencing any of those compoenents via reflection, we should be good with using only `dontwarn`.

electrostat: Perfect. I don't. I'll make the change.

Issue #188, opened 2018-07-23
Investigate NPE due to TelemetryService.obtainBoundInstances()


electrostat:
A non-replicable bug has been reported where an NPE is thrown when attempting to obtainBoundInstances. This is caused by when stopping the Telemetry Service.

Will investigate further and track down cause/solution.

lilykaiser:
Capturing that @zugaldia suggested internally to
> start with a null check when stopTelemetryService invokes obtainBoundInstances() as an initial workaround

PR #221, opened 2018-09-25
StopTelemetryService obtainBoundInstances bug fix

electrostat:
Under certain circumstances, `telemetryService` is null and causes a crash when calling `obtainBoundInstances()` within `stopTelemetryService`. Add in check to prevent crash.

Fixes https://github.com/mapbox/mapbox-events-android/issues/188


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -468,9 +468,11 @@ private void unbindTelemetryService() {
   private void stopTelemetryService() {
     TelemetryLocationEnabler.LocationState telemetryLocationState = telemetryLocationEnabler
```
andrlee: shouldn't null check be at this line? as well as should we escape early? 
`if (telemetryService == null) {
     /* log error */ 
     return;
  }
`

Issue #225, opened 2018-09-26
FileAttachments seems to be sent two times


yunikkk:
Using telemetry lib version 3.2.0 I noticed weird debugging logs about sending attachment events when `MapboxTelemetry.disable()` is called. 
Though, seems that attachments are already sent and callbacks `onAttachmentResponse` fired.

Digged a bit into a source code and looks that attachments are sent AND placed into the queue at the same time, so they later are sent one more time.

electrostat:
Looked into this. The `AttachmentEvent` is being added to the queue still, but the files are not re-uploaded. We are basically sending the metadata of the attachments 2x. To eliminate this, we should change how we handle `sendEventIfWhitelisted`. There is a missing `return true` line within the attachment if statement, which causes the queueing logic to still continue after the fact.

electrostat:
We also should add a unit test to check if whitelisted events are added to the queue. Will work on this tomorrow.

PR #226, opened 2018-10-02
Fix FileAttachment Queue Bug

electrostat:
Prevent `FileAttachment` event from being added to queue, in addition to sending it up through attachment logic.

Add in unit test to confirm that `whitelisted` events are not added to queue.\

Fixes https://github.com/mapbox/mapbox-events-android/issues/225

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
@@ -241,6 +241,10 @@ public class MapboxTelemetry implements FullQueueCallback, EventCallback, Servic
     return isLocationOpted;
   }
 
+  EventsQueue getQueue() {
+    return queue;
+  }
+
   private void startTelemetryService() {
     TelemetryLocationEnabler.LocationState telemetryLocationState = telemetryLocationEnabler
       .obtainTelemetryLocationState();
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
@@ -241,8 +241,8 @@ public class MapboxTelemetry implements FullQueueCallback, EventCallback, Servic
     return isLocationOpted;
   }
 
-  EventsQueue getQueue() {
-    return queue;
+  Boolean isQueueEmpty() {
+    return queue.queue.size() == 0;
   }
 
   private void startTelemetryService() {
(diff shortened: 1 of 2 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -241,6 +241,10 @@ boolean optLocationOut() {
     return isLocationOpted;
   }
 
+  EventsQueue getQueue() {
```
andrlee: what is the benefit of exposing/relaxing access restriction to queue object? it seems you just want to check if queue is empty, can we instead expose `boolean isQueueEmpty()` api?

libtelemetry/src/test/java/com/mapbox/android/telemetry/MapboxTelemetryTest.java
```diff
@@ -473,6 +473,38 @@ public void checksSessionIdRotationIntervalNotUpdated() throws Exception {
     assertFalse(notUpdatedSessionInterval);
   }
 
+  @Test
+  public void checksAppUserTurnstileNotQueued() throws Exception {
+    Context mockedContext = obtainNetworkConnectedMockedContext();
```
andrlee: is there a way to avoid mocking external classes (android framework and okhttp) in our tests?

libtelemetry/src/test/java/com/mapbox/android/telemetry/MapboxTelemetryTest.java
```diff
@@ -473,6 +473,38 @@ public void checksSessionIdRotationIntervalNotUpdated() throws Exception {
...
+    Callback mockedHttpCallback = mock(Callback.class);
+    EventsQueue eventsQueue = new EventsQueue(mock(FlushQueueCallback.class));
+    MapboxTelemetry theMapboxTelemetry = obtainMapboxTelemetryWith(mockedContext, mockedTelemetryClient,
+      mockedHttpCallback, eventsQueue);
+
+    Event whitelistedEvent = new AppUserTurnstile("anySdkIdentifier", "anySdkVersion", false);
+    theMapboxTelemetry.enable();
+    theMapboxTelemetry.push(whitelistedEvent);
+
+    assertEquals(0, theMapboxTelemetry.getQueue().queue.size());
```
andrlee: you could also do `assertTrue(theMapboxTelemetry.isQueueEmpty())` if you'd expose `boolean isQueueEmpty()` api.

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -241,6 +241,10 @@ boolean optLocationOut() {
     return isLocationOpted;
   }
 
+  Boolean isQueueEmpty() {
```
andrlee: i'd suggest to avoid boxed types and use primitives, in this particular case `boolean` is the way to go.

Issue #222, opened 2018-09-25
Add Exception for Missing FINE Location Permission


electrostat:
Add an exception that will be thrown when `ACCESS_FINE_LOCATION` permission is removed and before the `Passive` location engine is started by the service. Prevent potential crash and communicate issue to developer.

PR #227, opened 2018-10-02
ACCESS_FINE_LOCATION permission exception

electrostat:
Detect if `ACCESS_FINE_LOCATION` permission has been removed from the manifest and throw an exception.

Fixes https://github.com/mapbox/mapbox-events-android/issues/222

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
@@ -39,7 +39,6 @@ public class TelemetryService extends Service implements TelemetryCallback, Loca
   private LocationEnginePriority locationPriority = LocationEnginePriority.NO_POWER;
   private CopyOnWriteArraySet<ServiceTaskCallback> serviceTaskCallbacks = null;
   private TelemetryLocationEnabler telemetryLocationEnabler;
-  private Logger logger;
   // For testing only:
   private boolean isLocationEnablerFromPreferences = true;
   private boolean isLocationReceiverRegistered = false;
@@ -48,7 +47,6 @@ public class TelemetryService extends Service implements TelemetryCallback, Loca
   @Override
   public void onCreate() {
     super.onCreate();
-    this.logger = new Logger();
     createLocationReceiver();
     createTelemetryReceiver();
     createServiceTaskCallbacks();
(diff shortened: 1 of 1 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
```diff
@@ -287,8 +292,20 @@ private boolean locationPermissionCheck() {
...
+
       return coarsePermission == PackageManager.PERMISSION_GRANTED
         || finePermission == PackageManager.PERMISSION_GRANTED;
     }
   }
+
+  private void checkFinePermission(int finePermission) {
+    if (finePermission != PackageManager.PERMISSION_GRANTED) {
+      try {
+        throw new Exception(MISSING_FINE_PERMISSION);
```
andrlee: what's the benefit of throwing exception in try and catch block? should we just return false if fine permission is unavailable? does that make sense if we're currently using passive provider which is not going to work with coarse permission?

electrostat: Yeah that was the question I was asking earlier. If we wanted to prevent the crash (with the false return here) and keep the exception to figure out why the listener is not started or just throw the exception and allow the crash to force the change. I'll do the first one. Thanks for the confirmation.

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
```diff
@@ -287,8 +290,26 @@ private boolean locationPermissionCheck() {
...
+
       return coarsePermission == PackageManager.PERMISSION_GRANTED
         || finePermission == PackageManager.PERMISSION_GRANTED;
     }
   }
+
+  private boolean checkFinePermission(int finePermission) {
+    if (finePermission != PackageManager.PERMISSION_GRANTED) {
+      try {
+        throw new Exception(MISSING_FINE_PERMISSION);
```
andrlee: i'm still not sure why we need to throw exception catch and handle it in the same method? since we're basically stuck with passive provider would it make sense just to check if fine permissions are granted and if not let the user know that we can't collect location preferably via SecurityException? Thoughts?

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
```diff
@@ -287,8 +291,21 @@ private boolean locationPermissionCheck() {
       int finePermission = PermissionChecker.checkSelfPermission(MapboxTelemetry.applicationContext,
         Manifest.permission.ACCESS_FINE_LOCATION);
 
+      if (checkFinePermission(finePermission)) {
+        return false;
+      }
+
       return coarsePermission == PackageManager.PERMISSION_GRANTED
```
andrlee: does it actually make sense to check for coarse permission with passive provider at all?

PR #231, opened 2018-10-08
Telemetry Metric Event

electrostat:
Create and manage an internal metric event to track the performance of the telemetry system and find areas to make changes and improvements.


electrostat:
@andrlee Agreed. I'm still in progress and working on it. Gotta get the counters, event creation and queuing trigger, and tests in, before it'll be ready for review.

andrlee:
sorry just noticed `in progress` tag - looked too early. ;)

libtelemetry/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,7 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.mapbox.android.telemetry">
 
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
```
andrlee: should we consult with maps and nav sdk teams if this is going to be an issue?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -240,4 +247,26 @@ private String determineConfigEndpoint() {
...
+    MetricUtils.calculateTimeDiff(serverTime);
+  }
+
+  private long obtainServerTime(Response response) {
+    Headers headers = response.headers();
+
+    String dateString = headers.get("Date");
+
+    SimpleDateFormat dateFormat = new SimpleDateFormat(DATE_FORMAT);
+    Date serverDate = new Date();
```
andrlee: what's the benefit of instantiating date object here? if date is invalid would you rather throw exception?

libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationReceiver.java
```diff
@@ -66,4 +68,8 @@ private LocationMapper obtainLocationMapper() {
 
     return locationMapper;
   }
+
+  private void recordLastLocation(Location location) {
```
andrlee: this method can be declared static

libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationReceiver.java
```diff
@@ -43,6 +43,8 @@ private boolean sendEvent(Location location) {
       return false;
     }
 
+    recordLastLocation(location);
```
andrlee: is this the right place to record last location?

libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationReceiver.java
andrlee: also what is the benefit of wrapping a single utility call with method?

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClient.java
```diff
@@ -127,7 +151,13 @@ TelemetryClientSettings obtainSetting() {
     return setting;
   }
 
-  private void sendBatch(List<Event> batch, Callback callback) {
+  void updateFailedRequests(int code) {
+    if (code >= 400) {
```
andrlee: should we get have a constant defined for this number?

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClient.java
```diff
@@ -184,4 +230,56 @@ private RequestBody reverseMultiForm(MultipartBody.Builder builder) {
 
     return builder.build();
   }
+
+  private MetricEvent buildMetricEvent() {
```
andrlee: i'm confused by how responsibilities are divided between classes that are sharing counters logic. can you draw a quick class diagram? perhaps, it will help you see what't the problem is.

libtelemetry/src/test/java/com/mapbox/android/telemetry/TelemetryClientAppUserTurnstileEventTest.java
```diff
@@ -19,6 +21,8 @@
   @Test
   public void sendsTheCorrectBodyPostingAppUserTurnstileEvent() throws Exception {
     Context mockedContext = mock(Context.class, RETURNS_DEEP_STUBS);
+    final ConnectivityManager manager = mock(ConnectivityManager.class);
```
andrlee: do we plan to add more tests before we merge?

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClient.java
```diff
@@ -37,6 +40,18 @@
   private TelemetryClientSettings setting;
   private final Logger logger;
   private CertificateBlacklist certificateBlacklist;
+  private MetricUtils metricUtils;
```
andrlee: should we re-consider overloading TelemetryClient class with new fields and adhere to object oriented design principles? IMHO TelemetryClient is exposed to way too many implementation details of counters logic when it shouldn't. I'd rather see member class that hides all this complexity and exposes very simple API to the TelemetryClient class.

libtelemetry/src/main/java/com/mapbox/android/telemetry/MetricUtils.java
```diff
@@ -0,0 +1,134 @@
...
+
+  String getConfigResponse() {
+    return configResponse;
+  }
+
+  Location getLatestLocation() {
+    return latestLocation;
+  }
+
+  Map<String, Integer> calculateEventCountByType(List<Event> eventsQueue,
```
andrlee: should event count by type belong to event class?

libtelemetry/src/main/java/com/mapbox/android/telemetry/MetricUtils.java
```diff
@@ -0,0 +1,134 @@
...
+
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TimeZone;
+
+class MetricUtils {
```
andrlee: i'm a bit puzzled about the reasoning behind defining date as an instance variable and the rest being static variables. what is the purpose/responsibility of this class? if it is an utility shouldn't it be closed for instantiation? Now, classic utility class will have all static methods. This one does have a mix of both.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -226,6 +230,9 @@ private String determineConfigEndpoint() {
 
   private List<String> extractResponse(Response response) throws IOException {
     String responseData = response.body().string();
+    long serverTime = obtainServerTime(response);
```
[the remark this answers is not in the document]
electrostat: Goal was to prevent code duplication, since we are already accessing the response body string here. Saw no need to call it more than once.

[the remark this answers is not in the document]
andrlee: sorry, i might have not made myself very clear. my suggestion was to move the lines 233-235 to callback handler:  
```
 @Override
  public void onResponse(Call call, Response response) throws IOException {
    ResponseBody body = response.body();
    if (body == null) {
        // Log error
        return;
    }
    String responseData = body.string()
    // 1. Parse response
    List<String> revokedKeys = extractResponse(responseData);
    // 2. Setup metrics
    setupMetrics(obtainServerTime(response), responseData);
    // 3. Save revoked keys
    saveBlackList(revokedKeys);
  }
```

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -226,6 +230,9 @@ private String determineConfigEndpoint() {
 
   private List<String> extractResponse(Response response) throws IOException {
     String responseData = response.body().string();
```
andrlee: Btw, this line can result in NPE since body() can potentially return null.

PR #233, opened 2018-10-08
3.3.0 Telemetry Release

electrostat:
Releasing Telemetry 3.3.0

Fixes: https://github.com/mapbox/mapbox-events-android/issues/232

cc: @andrlee


andrlee:
@electrostat i think we should automate changelogs and version naming, so you don't have to do it manually.

PR #235, opened 2018-10-10
Update Load Event with sdkIdentifier and sdkVersion

electrostat:
Add optional `sdkIdentifier` and `sdkVersion` to `MapLoadEvent`

Addresses the immediate concerns of https://github.com/mapbox/mobile-telemetry/issues/239#event-1894099683 and stops currently failing test.


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapLoadEvent.java
```diff
@@ -124,6 +138,8 @@ public void writeToParcel(Parcel dest, int flags) {
     } else {
       dest.writeByte((byte) (wifi ? 0x01 : 0x00));
     }
+    dest.writeString(sdkIdentifier);
```
andrlee: shouldn't we write only when identifier is not null?

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapLoadEvent.java
```diff
@@ -88,6 +100,8 @@ private MapLoadEvent(Parcel in) {
     cellularNetworkType = in.readString();
     byte wifiVal = in.readByte();
     wifi = wifiVal == 0x02 ? null : wifiVal != 0x00;
+    sdkIdentifier = in.readString();
```
andrlee: would it be fair to assume that readString() on this line returns null if this field is not available?

Issue #234, opened 2018-10-09
Unexplained crashes related to Telemetry after upgrading to 6.5.0


malwoodsantoro:
Likely related to https://github.com/mapbox/mapbox-events-android/issues/146

Crashlytics report: 
![image](https://user-images.githubusercontent.com/19801577/46691614-29c60900-cbba-11e8-9d75-c50929f34776.png)

This appears to only be affecting users on Android 5.

malwoodsantoro:
Updated stack trace after updating SDK to 6.6.5 -

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Object android.content.Context.getSystemService(java.lang.String)' on a null object reference at com.mapbox.android.telemetry.TelemetryUtils.obtainApplicationState(TelemetryUtils.java:103)

at com.mapbox.android.telemetry.LocationEvent.<init>(LocationEvent.java:44)

at com.mapbox.android.telemetry.LocationMapper.createLocationEvent(LocationMapper.java:32)

at com.mapbox.android.telemetry.LocationMapper.from(LocationMapper.java:18)

at com.mapbox.android.telemetry.LocationReceiver.sendEvent(LocationReceiver.java:47)

at com.mapbox.android.telemetry.LocationReceiver.onReceive(LocationReceiver.java:26)

at android.support.v4.content.LocalBroadcastManager.executePendingBroadcasts(LocalBroadcastManager.java:313)

at android.support.v4.content.LocalBroadcastManager$1.handleMessage(LocalBroadcastManager.java:121)

at android.os.Handler.dispatchMessage(Handler.java:102)

at android.os.Looper.loop(Looper.java:148)

at android.app.ActivityThread.main(ActivityThread.java:5491)

at java.lang.reflect.Method.invoke(Method.java)

at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:728)

at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:618)

```

cc @electrostat @andrlee

andrlee:
fixed by https://github.com/mapbox/mapbox-events-android/pull/272

PR #241, opened 2018-10-16
PermissionChecker NPE

electrostat:
Check if `MapboxTelemetry.applicationContext` is `null` when checking location permissions.

Fixes https://github.com/mapbox/mapbox-events-android/issues/234

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
@@ -35,6 +35,8 @@ public class TelemetryService extends Service implements TelemetryCallback, Even
           + " the manifest. This is a required permission for Mapbox."
           + "Please add this permission back into your manifest, "
           + "so our system can work properly";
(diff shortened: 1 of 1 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
```diff
@@ -273,9 +273,12 @@ private boolean locationPermissionCheck() {
     if (Build.VERSION.SDK_INT >= API_LEVEL_23) {
       return PermissionsManager.areLocationPermissionsGranted(this);
     } else {
-      int finePermission = PermissionChecker.checkSelfPermission(MapboxTelemetry.applicationContext,
-              Manifest.permission.ACCESS_FINE_LOCATION);
-      return checkFinePermission(finePermission);
+      if (MapboxTelemetry.applicationContext != null) {
```
andrlee:
```suggestion
      if (MapboxTelemetry.applicationContext == null) {
         // Log error
         return false;
      }
```
Should we log error, such that we have a better understanding why permissions weren't granted?

PR #245, opened 2018-10-18
Add support for Vision object detection event (v2)

andrlee:
Note: Please refrain from using `VisionEventFactory` it is rather redundant and will be deprecated. 

/cc @l-r

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/VisionObjectDetectionEvent.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/VisionObjectDetectionEvent.java
@@ -145,22 +145,22 @@ public class VisionObjectDetectionEvent extends Event implements Parcelable {
 
   @Override
   Event.Type obtainType() {
-    return Event.Type.VIS_GENERAL;
+    return Type.VIS_OBJ_DETECTION;
   }
 
   private VisionObjectDetectionEvent(Parcel in) {
     this.event = in.readString();
     this.created = in.readString();
-    this.object_lat = in.readDouble();
-    this.object_lon = in.readDouble();
-    this.vehicle_lat = in.readDouble();
-    this.vehicle_lon = in.readDouble();
-    this.clazz = in.readString();
-    this.sign_value = in.readString();
-    this.object_size_width = in.readDouble();
-    this.object_size_height = in.readDouble();
-    this.object_pos_height = in.readDouble();
-    this.distance_from_camera = in.readDouble();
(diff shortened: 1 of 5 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/VisionObjectDetectionEvent.java
```diff
@@ -0,0 +1,212 @@
...
+    return distance_from_camera;
+  }
+
+  public void setDistanceFromCamera(double distance_from_camera) {
+    this.distance_from_camera = distance_from_camera;
+  }
+
+  @Override
+  Event.Type obtainType() {
+    return Event.Type.VIS_GENERAL;
```
yunikkk: VIS_OBJECT_DETECTION ?

libtelemetry/src/main/java/com/mapbox/android/telemetry/VisionObjectDetectionEvent.java
```diff
@@ -0,0 +1,212 @@
...
+  public int describeContents() {
+    return 0;
+  }
+
+  @Override
+  public void writeToParcel(Parcel dest, int flags) {
+    dest.writeString(event);
+    dest.writeString(created);
+
+    writeDoubleIfNotNull(dest, object_lat);
```
yunikkk: Will it work with reading then after writing fields optionally?

libtelemetry/src/main/java/com/mapbox/android/telemetry/VisionObjectDetectionEvent.java
andrlee: @yunikkk good catch! although i think the types are in general useless :)

libtelemetry/src/main/java/com/mapbox/android/telemetry/VisionObjectDetectionEvent.java
andrlee: you're right. it will make more sense to make use of binary flags - my bad. fixing it!

yunikkk: Or just writing default values - if that will be easier

andrlee: IMHO flag is a bit cleaner. it's hard sometimes to select default/invalid values for params like lat/lon.

PR #247, opened 2018-10-19
Mapbox Fused Location Engine

andrlee:
First attempt to improve user experience with MapboxFusedLocationEngine on devices where Google Play Services are not available. 

/cc @mapbox/android

### DIFF
--- a/liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
+++ b/liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
@@ -117,7 +117,9 @@ class MapboxFusedLocationEngineImpl extends AndroidLocationEngineImpl implements
       currentBestLocation = location;
     }
 
-    this.callback.onSuccess(LocationEngineResult.create(currentBestLocation));
+    if (callback != null) {
+      callback.onSuccess(LocationEngineResult.create(currentBestLocation));
+    }
   }
 
   @Override
--- a/liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
+++ b/liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
@@ -1,6 +1,8 @@
 package com.mapbox.android.core.location;
 
+import android.app.PendingIntent;
 import android.content.Context;
+import android.content.Intent;
 import android.location.Location;
 import android.location.LocationListener;
 import android.location.LocationManager;
@@ -10,8 +12,7 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.util.Log;
 
-import java.util.List;
-import java.util.concurrent.CopyOnWriteArrayList;
+import static com.mapbox.android.core.location.Utils.isBetterLocation;
 
 /**
  * Mapbox replacement for Google Play Services Fused Location Client
@@ -21,29 +22,25 @@ import java.util.concurrent.CopyOnWriteArrayList;
 class MapboxFusedLocationEngineImpl extends AndroidLocationEngineImpl implements LocationListener {
   private static final String TAG = "MapboxLocationEngine";
 
-  private static final int TWO_MINUTES = 1000 * 60 * 2;
-  private static final int ACCURACY_THRESHOLD_METERS = 200;
-
-  private final List<LocationEngineCallback<LocationEngineResult>> callbacks;
-
+  private LocationEngineCallback<LocationEngineResult> callback;
   private Location currentBestLocation;
 
   MapboxFusedLocationEngineImpl(@NonNull Context context) {
     super(context);
-    this.callbacks = new CopyOnWriteArrayList<>();
+    this.callback = null;
   }
 
   @NonNull
   @Override
   LocationListener createListener(final LocationEngineCallback<LocationEngineResult> callback) {
-    callbacks.add(callback);
+    this.callback = callback;
     return this;
   }
 
   @Nullable
   @Override
   public LocationListener removeLocationListener(@NonNull LocationEngineCallback<LocationEngineResult> callback) {
-    callbacks.remove(callback);
+    this.callback = null;
     return super.removeLocationListener(callback);
   }
 
@@ -64,11 +61,7 @@ class MapboxFusedLocationEngineImpl extends AndroidLocationEngineImpl implements
     super.requestLocationUpdates(request, listener, looper);
 
     // Start network provider along with gps
-    int priority = request.getPriority();
-    if ((priority == LocationEngineRequest.PRIORITY_HIGH_ACCURACY
-            || priority == LocationEngineRequest.PRIORITY_BALANCED_POWER_ACCURACY)
-            && currentProvider.equals(LocationManager.GPS_PROVIDER)) {
-
+    if (shouldStartNetworkProvider(request.getPriority())) {
       try {
         locationManager.requestLocationUpdates(LocationManager.NETWORK_PROVIDER,
                 request.getInterval(), request.getDisplacemnt(),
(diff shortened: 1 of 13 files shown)


andrlee:
@langsmith any chance we could test this while you in China? :)

langsmith:
Sure @andrlee , happy to check it out. cc @Chaoba 

Anything you'd like us to do in particular to test it?

andrlee:
@langsmith it would be ideal to use a snapshot of liblocation and implement a simple test app with it. the goal is to assess how reliably provider works in china in urban canyons.

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
```diff
@@ -0,0 +1,160 @@
...
+import android.location.LocationManager;
+import android.os.Bundle;
+import android.os.Looper;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.util.Log;
+
+import static com.mapbox.android.core.location.Utils.isBetterLocation;
+
+/**
```
yunikkk: I believe [LocationEngine](https://github.com/mapbox/mapbox-events-android/blob/de572d7f784141a5d57b65851e6fdce1ab3c9d32/liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineImpl.java) method `getLocationListener` is a typo and its actually `set...`.

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
```diff
@@ -0,0 +1,160 @@
...
+  private Location currentBestLocation;
+
+  MapboxFusedLocationEngineImpl(@NonNull Context context) {
+    super(context);
+    this.callback = null;
+  }
+
+  @NonNull
+  @Override
+  LocationListener createListener(final LocationEngineCallback<LocationEngineResult> callback) {
```
yunikkk: I'm a bit confused here... Why internal `createListener` is override and not `getLocationListener` to match overriden below `removeLocationListener`?

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
yunikkk: Also, should [AbstractLocationEngineImpl](https://github.com/mapbox/mapbox-events-android/blob/de572d7f784141a5d57b65851e6fdce1ab3c9d32/liblocation/src/main/java/com/mapbox/android/core/location/AbstractLocationEngineImpl.java) call `destroyListener` from `unmapLocationListener` when removing it from the list?

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  @Override
+  public void onProviderEnabled(String provider) {
+    Log.d(TAG, "onProviderEnabled: " + provider);
+  }
+
+  @Override
+  public void onProviderDisabled(String provider) {
+    Log.d(TAG, "onProviderEnabled: " + provider);
```
yunikkk: onProviderDisabled

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
andrlee: @yunikkk you actually got me thinking there - thanks for that 👍 I've found a way to get rig of destroy completely by moving removal logic upwards in the stack. hope it looks a bit cleaner.

yunikkk: Hmm and where I can find `remove` logic now? Dont see it in this PR...

yunikkk: Ahh, it's `removeLocationUpdates`, found it

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
```diff
@@ -0,0 +1,160 @@
...
+    if (shouldStartNetworkProvider(request.getPriority())) {
+      try {
+        locationManager.requestLocationUpdates(LocationManager.NETWORK_PROVIDER, request.getInterval(),
+                request.getDisplacemnt(), pendingIntent);
+      } catch (IllegalArgumentException iae) {
+        iae.printStackTrace();
+      }
+    }
+
+    // TODO: should we broadcast last location?
```
yunikkk: IMHO it would be good and consistent to have the same behavior with the previous one, which broadcasts it, too

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
```diff
@@ -0,0 +1,160 @@
...
+    } else {
+      callback.onFailure(new Exception("Last location unavailable"));
+    }
+  }
+
+  @Override
+  public void requestLocationUpdates(@NonNull LocationEngineRequest request,
+                                     @NonNull LocationListener listener,
+                                     @Nullable Looper looper) throws SecurityException {
+    super.requestLocationUpdates(request, listener, looper);
```
yunikkk: Could you please explain the logic behind possible `locationManager.requestLocationUpdates` here and in `super.` ?
Will we then request updates from locationManager multiple times if `shouldStartNetworkProvider` is true?

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
yunikkk: So here I was just surprised that it was destroyed from `map` be not from `unmap`, now it's clear

yunikkk: The follow up question is - how are `listeners` used here? First I thought that all the location updates will be sent to listeners but that seems not the case though...

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
andrlee: parent class is responsible for choosing the best provider available at a given moment and starting updates with it, the child class is responsible for figuring out whether these updates (in most cases gps) need to be supplemented with network location updates. if `shouldStartNetworkProvider == true` it is technically dispatches two parallel requests to both network and gps providers and the engine will get fused updates from both.

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
andrlee: agreed. this is intentional, as i'm not sure how `lastLocation` will impact user experience: my goal would be to run this engine both in background and foreground mode and analyze the diff in the experience.

liblocation/src/main/java/com/mapbox/android/core/location/MapboxFusedLocationEngineImpl.java
andrlee: i'm keeping ref to `callback` to report back `fused` location. not all updates will be sent to listeners potentially but only the ones that are qualified as `isBetterLocation` compared to last known location.

yunikkk: Yeah, `callback` is clear, I tried to understand what is the point to keep list of all callbacks in `listeners` field in [AbstractLocationEngine](https://github.com/mapbox/mapbox-events-android/blob/de572d7f784141a5d57b65851e6fdce1ab3c9d32/liblocation/src/main/java/com/mapbox/android/core/location/AbstractLocationEngineImpl.java)

andrlee: @yunikkk initially listeners field was created to keep track of unique `LocationEngineListener` callbacks, since technically user can create multiple requests and choose to handle updates from them with different callbacks, which could be mostly the case for Android location engine (not so much for Google Play version of fused provider), but since now we have our fused version of engine, this use case become less valuable. So you brought up a really good point, perhaps it'd make sense to simplify listener logic in near future.

PR #249, opened 2018-10-22
Unable to start unregistered service fix

electrostat:
`onServiceDisconnected` is not called following `unbindService` and can not be relied on for triggering certain behaviors. Remove functionality from this call and instead put that functionality after calling `unbindService` directly. 

Service is not reregistering, since service is shown as bound and not null when that is not the case.


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -556,6 +554,8 @@ private void stopLocation() {
   private boolean unbindServiceConnection() {
     if (TelemetryUtils.isServiceRunning(TelemetryService.class)) {
       applicationContext.unbindService(serviceConnection);
```
andrlee: is there a reason we're using both `startService/stopService` and `bindService/unbindService`? Btw, calling `unbindService` on the service that was started with `startService` may not actually stop the service until you call `service.stopSelf()`.

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -394,8 +394,6 @@ public void onServiceConnected(ComponentName className, IBinder service) {
 
       @Override
       public void onServiceDisconnected(ComponentName className) {
-        telemetryService = null;
```
andrlee: this is still required IMHO for the cases when connection with the service is unexpectedly interrupted.

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
andrlee: are we confident that this is not the root cause of the issue we're trying to solve? can you help me here to be a little less skeptical about the fix by adding unit test/instrumentation test?

andrlee:
@electrostat i'd follow this guideline: https://developer.android.com/training/testing/integration-testing/service-testing#java and implement basic integration test for `TelemeteryService`

andrlee:
@electrostat i'm proposing to share detailed root cause analysis of why `telemetryService == null` while clients are still actively interacting with it and have instrumentation test in place that proves that this is no longer an issue.

PR #254, opened 2018-10-24
Add lifecycle methods to track lifecycle of parent activity

electrostat:
Add onPause and onResume lifecycle methods to track parent activity lifecycle better. Remove old reliance on Lifecycle observer, which can cause a crash in certain circumstances.


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -28,8 +24,7 @@
 import okhttp3.Response;
 import okhttp3.ResponseBody;
 
-public class MapboxTelemetry implements FullQueueCallback, EventCallback, ServiceTaskCallback,
-  LifecycleObserver {
+public class MapboxTelemetry implements FullQueueCallback, EventCallback, ServiceTaskCallback {
```
andrlee: since you're removing ` LifecycleObserver` this change is breaking public APIs right?

electrostat: I actually will not have to do that. I have come up with a much better and effective solution. Just pulled in master to get the code coverage for planning out tests.

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -517,11 +546,11 @@ private PermissionCheckRunnable obtainPermissionCheckRunnable() {
   }
 
   private void startLocation() {
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-      if (ProcessLifecycleOwner.get().getLifecycle().getCurrentState().isAtLeast(Lifecycle.State.STARTED)) {
+    if (isLollipopOrHigher()) {
+      if (isAppInForeground()) {
         applicationContext.startService(obtainLocationServiceIntent());
```
andrlee: how about just catching `java.lang.IllegalStateException` here and on line 556? would that be sufficient to remedy the problem while we refactor out of `TelemetryService`?

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -497,11 +504,11 @@ private PermissionCheckRunnable obtainPermissionCheckRunnable() {
...
+    if (isLollipopOrHigher()) {
+      try {
         applicationContext.startService(obtainLocationServiceIntent());
-      } else {
-        ProcessLifecycleOwner.get().getLifecycle().addObserver(this);
+      } catch (IllegalStateException exception) {
+        logger.error(START_SERVICE_FAIL, exception.getMessage());
       }
     } else {
       applicationContext.startService(obtainLocationServiceIntent());
```
andrlee: this line can still throw `IllegalStateException`

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -65,6 +68,7 @@ public MapboxTelemetry(Context context, String accessToken, String userAgent) {
     initializeTelemetryListeners();
     initializeAttachmentListeners();
     initializeTelemetryLocationState(context.getApplicationContext());
+    this.logger = new Logger();
```
andrlee: why are we creating instance of the class that calls into static method of android `Log`? should we just simply use `Log` directly?

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -570,6 +583,7 @@ private Boolean checkNetworkAndParameters() {
 
   @OnLifecycleEvent(Lifecycle.Event.ON_START)
   void onEnterForeground() {
+    logger.error("test", "enter foreground");
```
andrlee: `test`? should we add proper tag?

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -182,6 +187,11 @@ public boolean removeAttachmentListener(AttachmentListener listener) {
     return attachmentListeners.remove(listener);
   }
 
+  @VisibleForTesting
+  void setBuildVersion(int buildVersion) {
```
andrlee: we don't actually need this setter method and `buildVersion` variable. we could instead stub `isLollipopOrHigher()` method instead.

PR #257, opened 2018-10-26
Handle potential 'null' last location returned by fused client

andrlee:
Minor pre-release bug fixes.


liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineResult.java
```diff
@@ -55,7 +55,7 @@ public static LocationEngineResult create(List<Location> locations) {
    * @since 3.0.0
    */
   public Location getLastLocation() {
```
yunikkk: nit: it would be great to add `@Nullable` annotation here.

andrlee: @yunikkk good catch! please continue keeping an eye on our PRs ;) I'll raise PR with the fix shortly.

Issue #263, opened 2018-10-29
Migrate TelemetryService to using getApplicationContext()


electrostat:
Address this work: https://github.com/mapbox/mapbox-events-android/issues/146#issuecomment-433968009

To prevent possible instances where `MapboxTelemetry.applicationContext` is null due to differing lifecycles, we will migrate from using this static context within our `TelemetryService` to using `getApplicationContext()`. This will prevent odd crashes we are noticing in over-managed OS's and under odd corner cases, where context is not set, but processes are running.

- [x] remove all reference to `MapboxTelemetry.applicationContext` from `TelemetryService`
- [x] replace with `getApplicationContext()`
- [x] remove context check method
- [x] add tests to confirm working fix

andrlee:
@electrostat i think the problem might be bigger than just invalid context ref, we need to get to the root cause: i.e. why context is null in a first place. You might need to instrument `TelemetryService` to get to that.

andrlee:
Fixed with https://github.com/mapbox/mapbox-events-android/pull/265

PR #265, opened 2018-10-29
Fix TelemetryService and applicationContext issue

electrostat:
Addresses: https://github.com/mapbox/mapbox-events-android/issues/263

Address issue where static applicationContext set by the activity is null, but service is still operating. Lifecycles are detached and can cause potential issues in lower API levels. Goal is to test and prevent crash from occurring. If possible, refactor all tangental code ran by the service to prevent this as well.


libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
```diff
@@ -181,6 +179,16 @@ boolean isTelemetryReceiverRegistered() {
     return isTelemetryReceiverRegistered;
   }
 
+  boolean locationPermissionCheck() {
```
andrlee: since we've dependency on `supportlib-v4` is there a reason why we're not simply doing:
```suggestion
 private boolean locationPermissionCheck() {
    return ContextCompat.checkSelfPermission(getApplicationContext(), Manifest.permission.ACCESS_FINE_LOCATION)
      == PackageManager.PERMISSION_GRANTED;
  }
```

andrlee: if we're changing method's visibility modifier in order to get it under test harness, we should always annotate it with `@VisibleForTesting` otherwise someone may think that we're not properly encapsulating our code. Now, another question is it really necessary to change its visibility in order to test it?

libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/TelemetryServiceTest.java
```diff
@@ -135,6 +139,17 @@ public void checksOnTaskRemovedCallbackWhenOnTaskRemovedCalled() throws Exceptio
     verify(mockedCallback, times(1)).onTaskRemoved();
   }
 
+  @Test
+  public void checksLocationPermission() throws Exception {
+    Intent serviceIntent = new Intent(InstrumentationRegistry.getTargetContext(), TelemetryService.class);
```
andrlee: what are we trying to test here for? since permission is already granted with `GrantPermissionRule` how useful is it? Should we consider implementing a test when permissions are not present and try to verify the outcome/result with public APIs exposed by `TelemetryService`?

gradle/dependencies.gradle
```diff
@@ -64,7 +64,11 @@ ext {
             junit                  : "junit:junit:${version.junit}",
             mockito                : "org.mockito:mockito-core:${version.mockito}",
             mockitoAndroid         : "org.mockito:mockito-android:${version.mockito}",
-            assertJ                : "org.assertj:assertj-core:${version.assertJ}"
+            assertJ                : "org.assertj:assertj-core:${version.assertJ}",
+
+            //AndroidX
+            androidXEspresso       : "androidx.test.espresso:espresso-core:${version.espressoVersion}",
```
andrlee: are these dependencies necessary to compile test code?

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryService.java
```diff
@@ -181,6 +178,12 @@ boolean isTelemetryReceiverRegistered() {
     return isTelemetryReceiverRegistered;
   }
 
+  @VisibleForTesting
+  boolean locationPermissionCheck() {
+    return ContextCompat.checkSelfPermission(getApplicationContext(), Manifest.permission.ACCESS_FINE_LOCATION)
```
andrlee: would you be able to test this fix on at API 19, 23, 26, 28?

PR #266, opened 2018-10-30
[WIP] Active location data collection

andrlee:
The location engine controller is implemented as state machine with effects, where states and events are modeled by a immutable value type objects. `handleEvent` method of each state class clearly describes all state transitions. `LocationEngineController`is the shell that hides the enums and value types, delegates state transitions and connects controller's interfaces.


alfwatt:
Stale

Issue #230, opened 2018-10-06
Crash on Samsung devices from AlarmManager


KairisCharm:
I recently started a beta release of a new version of my app that uses MapBox. 2 of our beta testers have reported this crash, one using a Samsung Galaxy S8, and the other using a Samsung Galaxy Note8, both running Android 8.0. 

> java.lang.ExceptionInInitializerError: 
  at com.mapbox.mapboxsdk.maps.Telemetry.obtainTelemetry (Telemetry.java:73)
  at com.mapbox.mapboxsdk.maps.Telemetry.initialize (Telemetry.java:30)
  at com.mapbox.mapboxsdk.Mapbox.initializeTelemetry (Mapbox.java:119)
  at com.mapbox.mapboxsdk.Mapbox.getInstance (Mapbox.java:49)
  at org.myapp.app.Application.onCreate (Application.kt:31)
  at android.app.Instrumentation.callApplicationOnCreate (Instrumentation.java:1125)
  at android.app.ActivityThread.handleBindApplication (ActivityThread.java:6056)
  at android.app.ActivityThread.-wrap1 (Unknown Source)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1764)
  at android.os.Handler.dispatchMessage (Handler.java:105)
  at android.os.Looper.loop (Looper.java:164)
  at android.app.ActivityThread.main (ActivityThread.java:6938)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.Zygote$MethodAndArgsCaller.run (Zygote.java:327)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1374)
Caused by: java.lang.SecurityException: 
  at android.os.Parcel.readException (Parcel.java:1959)
  at android.os.Parcel.readException (Parcel.java:1905)
  at android.app.IAlarmManager$Stub$Proxy.set (IAlarmManager.java:217)
  at android.app.AlarmManager.setImpl (AlarmManager.java:697)
  at android.app.AlarmManager.setInexactRepeating (AlarmManager.java:798)
  at com.mapbox.android.telemetry.AlarmSchedulerFlusher.schedule (AlarmSchedulerFlusher.java:40)
  at com.mapbox.android.telemetry.MapboxTelemetry.startAlarm (MapboxTelemetry.java:511)
  at com.mapbox.android.telemetry.MapboxTelemetry.startTelemetry (MapboxTelemetry.java:467)
[...]

PR #267, opened 2018-10-31
Update AlarmManager flag

electrostat:
Addresses: https://github.com/mapbox/mapbox-events-android/issues/230

On Samsung devices, the Alarms are not properly cancelled when requested. They fill up until the a `500` level threshold and cause a crash. Instead of cancelling, we are switching to updating the current alarm. 

Research seen here: https://stackoverflow.com/questions/29344971/java-lang-securityexception-too-many-alarms-500-registered-from-pid-10790-u


libtelemetry/src/test/java/com/mapbox/android/telemetry/AlarmSchedulerFlusherTest.java
```diff
@@ -67,4 +67,32 @@ public void checksAlarmUnregistered() throws Exception {
     verify(mockedAlarmManager, times(1)).cancel(eq(theAlarmSchedulerFlusher.obtainPendingIntent()));
     verify(mockedContext, times(1)).unregisterReceiver(mockedAlarmReceiver);
   }
+
+  @Test
+  public void checksAlarmUpdated() {
```
andrlee: looks like this test is verifying behavior of the `AlarmSchedulerFlusher` which is great, but what about `AlarmManager` itself? should we implement instrumentation test to make sure android framework is reacting properly to our change and there's no side effects of this change?

libtelemetry/src/main/java/com/mapbox/android/telemetry/AlarmSchedulerFlusher.java
```diff
@@ -28,7 +28,7 @@
   @Override
   public void register() {
     Intent alarmIntent = receiver.supplyIntent(requestCode);
-    pendingIntent = PendingIntent.getBroadcast(context, requestCode, alarmIntent, FLAG_CANCEL_CURRENT);
+    pendingIntent = PendingIntent.getBroadcast(context, requestCode, alarmIntent, FLAG_UPDATE_CURRENT);
```
andrlee: i'm looking at this change closely and after reading api docs i'm not sure why we need `FLAG_UPDATE_CURRENT` at all? We don't seem to be updating any extras with this pending intent...do you have any insights into why would it be potentially beneficial?

electrostat: The issue is that `FLAG_CANCEL_CURRENT` does not actually cancel the alarm on Samsung devices. What we do with this flag is it cancels the current alarm (if it exists) and then we update it. Update allows us to replace the alarm. If we just use no flag (0), we will have the same issue on Samsung and other devices now. We will keep generating new alarms and not getting rid of old ones. This way we are saying to the system to use only one alarm, not more. 

A possible alternative is to detect if a duplicate pending intent has been created using `NO_CREATE`flag and if so, delete that old one and then generate a new pending intent manually instead of letting the system handle it.

libtelemetry/src/main/java/com/mapbox/android/telemetry/AlarmSchedulerFlusher.java
```diff
@@ -1,35 +1,34 @@
...
   }
 
   @Override
   public void register() {
-    Intent alarmIntent = receiver.supplyIntent(requestCode);
-    pendingIntent = PendingIntent.getBroadcast(context, requestCode, alarmIntent, FLAG_CANCEL_CURRENT);
-    String action = SCHEDULER_FLUSHER_INTENT + Integer.toString(requestCode);
+    Intent alarmIntent = receiver.supplyIntent();
+    pendingIntent = PendingIntent.getBroadcast(context, REQUEST_CODE, alarmIntent, 0);
+    String action = SCHEDULER_FLUSHER_INTENT;
```
andrlee: should we just do `IntentFilter filter = new IntentFilter(SCHEDULER_FLUSHER_INTENT);` ?

libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/AlarmMangerInstrumentationTest.java
```diff
@@ -0,0 +1,60 @@
...
+    long elapsedMockedTime2 = 5000;
+
+    theAlarmSchedulerFlusher.register();
+    theAlarmSchedulerFlusher.scheduleExact(elapsedMockedTime);
+
+    theAlarmSchedulerFlusher.register();
+    theAlarmSchedulerFlusher.scheduleExact(elapsedMockedTime2);
+
+    try {
+      Thread.sleep(30000);
```
andrlee: this doesn't look like the most efficient way of testing async code. my recommendation would be to make use of combination of counting semaphore and atomic reference. check the example of similar test here: https://github.com/mapbox/mapbox-events-android/blob/ca8b49d00d42cec753cd36cf54b04b310fe04e77/liblocation/src/test/java/com/mapbox/android/core/location/LocationEngineTest.java#L40

libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/AlarmMangerInstrumentationTest.java
```diff
@@ -0,0 +1,63 @@
...
+      @Override
+      public void onPeriodRaised() {}
+
+      @Override
+      public void onError() {}
+    }) {
+      @Override
+      public void onReceive(Context context, Intent intent) {
+        super.onReceive(context, intent);
+        broadcastTrack.set(increment + 1);
```
andrlee: How about doing something like this? 
```
int count = broadcastTrack.get() == null ? 0 : broadcastTrack.get().value();
broadcastTrack.set(count + 1); 
```
then you don't need to maintain static variable

libtelemetry/src/main/java/com/mapbox/android/telemetry/AlarmReceiver.java
```diff
@@ -20,14 +21,20 @@
   @Override
   public void onReceive(Context context, Intent intent) {
     String alarm = intent.getStringExtra(ALARM_FIRED_INTENT_KEY);
-    if (ON_ALARM_INTENT_EXTRA.equals(alarm)) {
+    int requestCode = intent.getIntExtra(REQUEST_CODE_EXTRA, 0);
```
andrlee: `if (SCHEDULER_FLUSHER_INTENT.equals(intent.getAction()) {} `

PR #274, opened 2018-11-03
Optimise ConcurrentQueue

Chaoba:
Optimise ConcurrentQueue and add test cases


libtelemetry/src/main/java/com/mapbox/android/telemetry/ConcurrentQueue.java
```diff
@@ -8,35 +8,32 @@
 
 class ConcurrentQueue<T> {
```
andrlee: IMHO this class should be put to bed completely, i don't see a reason in maintaining additional generic wrapper for `ConcurrentLinkedQueue`

libtelemetry/src/main/java/com/mapbox/android/telemetry/FlushQueueCallback.java
```diff
@@ -3,5 +3,5 @@
 
 interface FlushQueueCallback {
```
andrlee: same here, i think this interface is redundant, should be put to bed as well.

Chaoba: Yeah, I have the same thought when I read here

Chaoba:
Close since we will refactor this part.

Issue #291, opened 2018-11-20
User Agent is not set up for Vision SDK Attachments. New User Agent needed 


BenDenen:
We are testing Vision SDK and found out that we can not set up right User Agent for Telemetry SDK. There are 2 issues:
1) We need new User Agent (`MapboxVision`) in list of supported Agents
[Link to the code](https://github.com/mapbox/mapbox-events-android/blob/e7b05b1ab42b3bf01bba6c3517b7ec15d6521393/libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java#L39)

2) Telemetry Client does not set up User Agent when it sent Attachment events
[Link to the code](https://github.com/mapbox/mapbox-events-android/blob/e7b05b1ab42b3bf01bba6c3517b7ec15d6521393/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClient.java#L96)

PR #292, opened 2018-11-20
Add Vision UserAgent

electrostat:
Add custom user agent for the Vision SDK. Additionally, add user agent when Attachment Events are sent through the system. Currently omitted. 

Fixes https://github.com/mapbox/mapbox-events-android/issues/291


libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -36,6 +36,7 @@
   private static final String NAVIGATION_USER_AGENT = "mapbox-navigation-android/";
   private static final String NAVIGATION_UI_USER_AGENT = "mapbox-navigation-ui-android/";
   private static final String MAPS_USER_AGENT = "mapbox-maps-android/";
+  private static final String VISION_USER_AGENT = "MapboxVision/";
```
andrlee: should we be consistent with user agent format? why not `mapbox-vision-android`?

electrostat: This was the requested change from stakeholder (@BenDenen) and matches the iOS implementation, per slack discussion with @l-r. Additionally, we don't actually have a set format, with `MapboxEventsAndroid`, `MapboxTelemetryAndroid`, and `MapboxEventsUnityAndroid` all being valid user agents. 

The tweak I would think to make, if it is ok with everyone involved, is to either add the hyphen system (suggested above) or add `Android` to end of `MapboxVision` (MapboxVisionAndroid) to help differentiate between iOS and android implementations. If that seems unnecessary, since we send back platform with events, then I am fine with how it is now.

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClient.java
```diff
@@ -92,6 +92,7 @@ void sendAttachment(Attachment attachment, final CopyOnWriteArraySet<AttachmentL
```
BenDenen: Thank for changes. Is it possible to add extra logs here as we did for `sendBatch` ?
I mean something like 
`if (isExtraDebuggingNeeded()) { 
     logger.debug(LOG_TAG, String.format(Locale.US, EXTRA_DEBUGGING_LOG, url, visionAttachments.size(), userAgent, metadataList));	
}`

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
andrlee: > Additionally, we don't actually have a set format, with MapboxEventsAndroid, MapboxTelemetryAndroid, and MapboxEventsUnityAndroid all being valid user agents.

If that's true, what's the point in maintaining `VALID_USER_AGENTS`? And would it make more sense to pass through network error if `userAgent` field is invalid (provided that backend will return error code for invalid `userAgent`)?

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -169,9 +153,7 @@ public void updateDebugLoggingEnabled(boolean isDebugLoggingEnabled) {
   }
 
   public void updateUserAgent(String userAgent) {
-    if (isUserAgentValid(userAgent)) {
```
andrlee: should we keep this check? if `userAgent` is empty or `null` the next line will produce crash.

andrlee:
@electrostat should we also implement integration test and test with staging environment?

PR #297, opened 2018-11-26
Add UserAgent header to Config Request

electrostat:
Fixes https://github.com/mapbox/mapbox-events-android/issues/296

Add UserAgent header to config request.

PR #299, opened 2018-11-27
Handle IndexOutOfBounds for Cert Blacklist retrieval

snkashis:
I saw this come up in our Crashlytics logs after upgrading to the most recent version. Doesn't seem like it needs to be a fatal.

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
@@ -80,7 +80,7 @@ class CertificateBlacklist implements Callback {
         try {
           blacklist = obtainBlacklistContents(file);
           blacklist.remove(BLACKLIST_HEAD);
-        } catch (IOException exception) {
+        } catch (IOException | IndexOutOfBoundsException exception) {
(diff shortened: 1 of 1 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -80,7 +80,7 @@
         try {
           blacklist = obtainBlacklistContents(file);
           blacklist.remove(BLACKLIST_HEAD);
-        } catch (IOException exception) {
+        } catch (IOException | IndexOutOfBoundsException exception) {
```
andrlee: @electrostat can you please look into why `blacklist.remove(BLACKLIST_HEAD)` throws `IOBE`?

andrlee:
@snkashis thanks a lot for submitting this change! it looks good to me, we'd like to dig a bit deeper though into why this exception occurs in a first place.

electrostat:
@snkashis this change was included within https://github.com/mapbox/mapbox-events-android/pull/297 while we work on finding a solution to the root problem.

Going to close this PR and move all discussion to this issue: https://github.com/mapbox/mapbox-events-android/issues/307

Issue #307, opened 2018-11-30
IndexOutOfBounds Exception when getting or removing BLACKLIST_HEADER


electrostat:
Find root cause of out of bounds exception when parsing Blacklist file. Seems to be related to no data being within Blacklist File, yet file exists.

Initial bug report: https://github.com/mapbox/mapbox-events-android/issues/307#issue-386357114

snkashis:
So far, this has triggered on two devices, a Moto G (2nd Gen) (Marshmellow), and a TRT-LX3(Nougat)

```
java.util.ArrayList.throwIndexOutOfBoundsException (ArrayList.java:255)
  | java.util.ArrayList.remove (ArrayList.java:403)
  | com.mapbox.android.telemetry.CertificateBlacklist.retrieveBlackList (CertificateBlacklist.java:82)
  | com.mapbox.android.telemetry.CertificatePinnerFactory.removeBlacklistedPins (CertificatePinnerFactory.java:51)
  | com.mapbox.android.telemetry.CertificatePinnerFactory.provideCertificatePinnerFor (CertificatePinnerFactory.java:31)
  | com.mapbox.android.telemetry.TelemetryClientSettings.configureHttpClient (TelemetryClientSettings.java:151)
  | com.mapbox.android.telemetry.TelemetryClientSettings.getClient (TelemetryClientSettings.java:54)
  | com.mapbox.android.telemetry.TelemetryClient.sendBatch (TelemetryClient.java:150)
  | com.mapbox.android.telemetry.TelemetryClient.sendEvents (TelemetryClient.java:61)
  | com.mapbox.android.telemetry.MapboxTelemetry.sendEvents (MapboxTelemetry.java:395)
  | com.mapbox.android.telemetry.MapboxTelemetry.sendEventsIfPossible (MapboxTelemetry.java:371)
  | com.mapbox.android.telemetry.MapboxTelemetry.sendEventIfWhitelisted (MapboxTelemetry.java:493)
  | com.mapbox.android.telemetry.MapboxTelemetry.push (MapboxTelemetry.java:141)
  | com.mapbox.mapboxsdk.module.telemetry.TelemetryImpl.onAppUserTurnstileEvent (TelemetryImpl.java:61)
  | com.mapbox.mapboxsdk.maps.MapView.onCreate (MapView.java:286)

```

electrostat:
@snkashis thanks for the extra info. This will help in finding the heart of the issue.

electrostat:
[...]

PR #311, opened 2018-12-04
Refactor Certificate Blacklist

electrostat:
A refactor how we implement requesting, saving, and reading our blacklist of revoked keys.

Goal is to use fewer file apis, reduce work during requests, and simplify 24 hr checkin, while avoiding read/write collisions.

Fixes: https://github.com/mapbox/mapbox-events-android/issues/307#issue-386357114


libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -58,32 +60,37 @@
...
   private final OkHttpClient client;
+  private final SharedPreferences sharedPreferences;
+  private List<String> revokedKeys;
 
   CertificateBlacklist(Context context, String accessToken, String userAgent, OkHttpClient client) {
     this.context = context;
     this.accessToken = accessToken;
     this.userAgent = userAgent;
     this.client = client;
+    this.sharedPreferences = TelemetryUtils.obtainSharedPreferences(context);
```
andrlee: why not pass `SharedPreferences` ref as constructor param?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificatePinnerFactory.java
```diff
@@ -48,7 +48,7 @@ private void addCertificatesPins(Map<String, List<String>> pins, CertificatePinn
 
   private Map<String, List<String>> removeBlacklistedPins(Map<String, List<String>> pins,
                                                           CertificateBlacklist certificateBlacklist) {
-    List<String> blackList = certificateBlacklist.retrieveBlackList();
+    List<String> blackList = certificateBlacklist.getRevokedKeys();
```
andrlee: this line is still called with every request, do we actually need to do this? should this be only called every 24 hours when blacklist is potentially updated?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -223,20 +209,12 @@ private String determineConfigEndpoint() {
...
-    Gson gson = new Gson();
-    JsonObject jsonObject = gson.fromJson(responseData, JsonObject.class);
-
-    Type listType = new TypeToken<List<String>>() {}.getType();
-    List revokedKeys = gson.fromJson(jsonObject.get("RevokedCertKeys"), listType);
-
-    return revokedKeys;
-  }
+  private void saveTimestamp() {
+    Date date = new Date();
```
andrlee: what's up with `Date` here vs. `System.currentTimeMillis()` on line 95? Should we use `Date` there as well?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -126,13 +118,17 @@ HttpUrl generateRequestUrl() {
       .build();
   }
 
-  private void saveBlackList(List<String> revokedKeys) {
-    String fileContents = createListContent(revokedKeys);
+  private void saveBlackList(Response response) throws IOException {
+    Gson gson = new GsonBuilder().create();
+    JsonObject responseJson = gson.fromJson(response.body().string(), JsonObject.class);
```
andrlee: on this line `response.body()` can return `null` so we may end up with `NPE`

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -126,13 +118,17 @@ HttpUrl generateRequestUrl() {
...
+
+    responseJson = addSHAtoKeys(responseJson);
     FileOutputStream outputStream = null;
+    saveTimestamp();
 
     try {
       outputStream = context.openFileOutput(BLACKLIST_FILE, Context.MODE_PRIVATE);
-      outputStream.write(fileContents.getBytes());
+      outputStream.write(responseJson.toString().getBytes());
```
andrlee: just wondering whether we should encrypt this file?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -151,54 +147,44 @@ public void onFailure(Call call, IOException exception) {
...
+    Gson gson = new Gson();
 
-    StringBuilder content = new StringBuilder(date.getTime() + NEW_LINE);
+    JsonArray jsonArray = jsonObject.getAsJsonArray("RevokedCertKeys");
+    JsonArray shaKeys = new JsonArray();
 
-    for (String key : revokedKeys) {
-      content.append(SHA256).append(key).append(NEW_LINE);
+    for (JsonElement key : jsonArray) {
+      shaKeys.add(SHA256 + key.getAsString());
```
andrlee: why are we doing this? is this how server wants them to be formatted and then why blacklisted hashes aren't sent to us in the right format?

electrostat: Correct on both accounts. This is how the hashes need to be formatted and they are sent to us without this formatting. We had a discussion about it at the time, but I do not recall the reason this wasn't done. We can make a request to the backend team to get this change made to make everything easier on us.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -58,32 +60,37 @@
   private final String accessToken;
   private final String userAgent;
   private final OkHttpClient client;
+  private final SharedPreferences sharedPreferences;
+  private List<String> revokedKeys;
```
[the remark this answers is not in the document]
electrostat: This variables is updated below within `addSHAtoKeys()` so that we are updating the keys immediately, so there isn't a 24hr delay on getting the latest revokedKeys to be used by the system. It should stay up-to-date and within a state of flux. This is why it was not made a final variable. Do you have recommendations for changing how this works?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -223,20 +210,11 @@ private String determineConfigEndpoint() {
...
-  }
+  private void saveTimestamp() {
+    Date date = new Date();
+    SharedPreferences.Editor editor = sharedPreferences.edit();
 
-  private String[] separateUrlSegments(String url) {
-    url = url.replace(HTTPS, EMPTY_STRING);
-    return url.split(BACKSLASH);
+    editor.putLong(MAPBOX_SHARED_PREFERENCE_KEY_BLACKLIST_TIMESTAMP, date.getTime());
+    editor.apply();
```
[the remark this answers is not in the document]
electrostat: This is done since `apply` is async (writes to memory immediately and schedules write to long term storage) and keeps the process off the main thread. We generally don't use `commit` for this very reason. We can discuss changing that methodology and investigating ways to handle this.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
electrostat: What would the security concern be over anyone knowing blacklisted keys? We already have all our keys viewable within the code itself. We also save this as `MODE_PRIVATE` which keeps it secure to the app. I believe this was acceptable to the security team during development of this feature.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -223,20 +210,28 @@ private String determineConfigEndpoint() {
...
-  private String[] separateUrlSegments(String url) {
-    url = url.replace(HTTPS, EMPTY_STRING);
-    return url.split(BACKSLASH);
+  private boolean isFileOpened(File file) throws IOException {
+    boolean fileOpen = false;
+    FileChannel channel = new RandomAccessFile(file, "rw").getChannel();
+    FileLock lock = channel.lock();
+
+    try {
+      lock = channel.tryLock();
```
andrlee: would this be called on ui thread?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -142,63 +133,59 @@ private void saveBlackList(List<String> revokedKeys) {
...
+    Gson gson = new Gson();
 
-    StringBuilder content = new StringBuilder(date.getTime() + NEW_LINE);
+    JsonArray jsonArray = jsonObject.getAsJsonArray("RevokedCertKeys");
+    JsonArray shaKeys = new JsonArray();
 
-    for (String key : revokedKeys) {
-      content.append(SHA256).append(key).append(NEW_LINE);
+    for (JsonElement key : jsonArray) {
+      shaKeys.add(SHA256 + key.getAsString());
```
andrlee: what if we send stripped version of the key to backend? also if they don't accept it should we file a ticket for them to either fix config or accept stripped version of the key?

libtelemetry/src/main/java/com/mapbox/android/telemetry/BlacklistClient.java
```diff
@@ -0,0 +1,32 @@
+package com.mapbox.android.telemetry;
+
+import android.util.Log;
+
+import okhttp3.Callback;
+import okhttp3.HttpUrl;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+
+class BlacklistClient {
```
andrlee: what's the benefit of wrapping a network request into a separate class while okhttp classes are still leaked into its owner?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -223,20 +210,28 @@ private String determineConfigEndpoint() {
...
+    editor.apply();
   }
 
-  private String[] separateUrlSegments(String url) {
-    url = url.replace(HTTPS, EMPTY_STRING);
-    return url.split(BACKSLASH);
+  private boolean isFileOpened(File file) throws IOException {
+    boolean fileOpen = false;
+    FileChannel channel = new RandomAccessFile(file, "rw").getChannel();
+    FileLock lock = channel.lock();
```
andrlee: would this be called on ui thread?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -56,66 +61,49 @@
...
 
     if (directory.isDirectory()) {
       File file = new File(directory, BLACKLIST_FILE);
 
       if (file.exists()) {
         try {
-          blacklist = obtainBlacklistContents(file);
-          blacklist.remove(BLACKLIST_HEAD);
-        } catch (IOException | IndexOutOfBoundsException exception) {
+          if (!isFileOpened(file)) {
```
andrlee: is there a lock free way to make sure we don't need to worry about at any time that file is read/written at the same time?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -142,63 +133,59 @@ private void saveBlackList(List<String> revokedKeys) {
...
   @Override
   public void onResponse(Call call, Response response) throws IOException {
-    List<String> revokedKeys = extractResponse(response);
-
-    saveBlackList(revokedKeys);
+    Log.e("test", "onResponse: " + response);
+    if (response.body() != null) {
+      saveBlackList(response.body());
+    }
   }
```
andrlee: should we call `saveTimestamp()` here instead?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -142,63 +133,59 @@ private void saveBlackList(List<String> revokedKeys) {
...
+  private JsonObject addSHAtoKeys(JsonObject jsonObject) {
+    Gson gson = new Gson();
 
-    StringBuilder content = new StringBuilder(date.getTime() + NEW_LINE);
+    JsonArray jsonArray = jsonObject.getAsJsonArray("RevokedCertKeys");
+    JsonArray shaKeys = new JsonArray();
 
-    for (String key : revokedKeys) {
-      content.append(SHA256).append(key).append(NEW_LINE);
+    for (JsonElement key : jsonArray) {
```
andrlee: this is also potential `NPE` if `jsonArray` is` null`, which is a possibility.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
electrostat: yeah we can use boolean within the class to do the checking and use it to control read/write.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -18,225 +22,166 @@
...
       try {
         outputStream.close();
       } catch (IOException exception) {
         Log.e(LOG_TAG, exception.getMessage());
+        fileOpen = false;
       }
     }
+
+    fileOpen = false;
+    saveTimestamp();
```
andrlee: redundant `saveTimestamp` call

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -18,225 +22,166 @@
...
-    return new HttpUrl.Builder().scheme(HTTPS_SCHEME)
-      .host(determineConfigEndpoint())
-      .addPathSegment(EVENT_CONFIG_SEGMENT)
-      .addQueryParameter(ACCESS_TOKEN_QUERY_PARAMETER, accessToken)
-      .build();
-  }
+  private void saveBlackList(ResponseBody responseBody) throws IOException {
+    fileOpen = true;
+    Gson gson = new GsonBuilder().create();
+    JsonObject responseJson = gson.fromJson(responseBody.string(), JsonObject.class);
```
andrlee: if `responseBody` is `null` this line will crash, if `responseBody.string()` returns empty string this will throw `JsonSyntaxException` which should be handled.

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -18,225 +22,166 @@
...
 
-  HttpUrl generateRequestUrl() {
-    return new HttpUrl.Builder().scheme(HTTPS_SCHEME)
-      .host(determineConfigEndpoint())
-      .addPathSegment(EVENT_CONFIG_SEGMENT)
-      .addQueryParameter(ACCESS_TOKEN_QUERY_PARAMETER, accessToken)
-      .build();
-  }
+  private void saveBlackList(ResponseBody responseBody) throws IOException {
+    fileOpen = true;
```
andrlee: this doesn't look like a viable solution for the problem. why we need this at all?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -18,126 +21,100 @@
...
     this.accessToken = accessToken;
-    this.userAgent = userAgent;
-    this.client = client;
+    this.blacklistClient = new BlacklistClient(userAgent, new OkHttpClient(), this);
+    retrieveBlackList();
   }
 
-  List<String> retrieveBlackList() {
+  boolean isBlacklisted(String hash) {
+    for (String revokedKey : revokedKeys) {
```
andrlee: `List<T>` has `contains()` method, should we just use it?

electrostat: Due to there being only a single instance of `CERTIFICATE_PINS` within the `CertificatePinnerFactory` class, we can not use a simple `contains()` method. Our Certificates need to keep their `shar256/` prefix, because being added after the `isBlacklisted` check will cause this prefix to be added on each additional request.  (i.e. sha256/sha256/sha256/...) 

We thus have a system where the local hashes have `sha256/` while the save blacklisted ones will not have this prefix. (since it was requested to be removed above). The current solution works within these parameters. An alternative may be switching around the analysis and seeing if any of the blacklisted hashes match the current hashes, but this would go back to our system to before this PR where we are returning a list of hashes.

andrlee: aren't we expecting `isBlacklisted` method to be called just once and not with every network request?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -18,225 +20,100 @@
...
+    } else {
+      retrieveBlackList(context.getFilesDir());
+    }
   }
 
-  List<String> retrieveBlackList() {
-    File directory = context.getFilesDir();
-    List<String> blacklist = new ArrayList<>();
+  boolean isBlacklisted(String hash) {
+    return revokedKeys.contains(hash);
```
andrlee: i still don't understand why this code is problematic..

libtelemetry/src/test/java/com/mapbox/android/telemetry/ConfigurationClientTest.java
```diff
@@ -0,0 +1,112 @@
...
+    this.configuationClient = new ConfiguationClient(mockedContext, TelemetryUtils.createFullUserAgent("AnUserAgent",
+      mockedContext), "anAccessToken", client);
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    server.shutdown();
+  }
+
+  @Test
```
andrlee: just one method?

PR #313, opened 2018-12-10
Resolve crash caused by removing items from empty List

Chaoba:
blacklist could be empty if there is an exception in `obtainBlacklistContents` method, so add a check to fix it.


Chaoba:
@electrostat Glad to hear about that, close this PR.

PR #318, opened 2018-12-21
Enable event creation outside of telem sdk

andrlee:
Remove abstract `obtainType` to enable event creation outside of telemetry sdk. This change is done with a focus on preserving current telem sdk api signatures.

An open question: since we don't have any events outside of telem sdk yet, should we remove `Parcelable` dependency before we expose it to public? Can we do it without doing a major release?

/cc @mapbox/android


andrlee:
It looks like this solution is the most optimal way to enable event creation outside of telem sdk. introducing a new interface for events will result in higher effort required to enable this feature for the most simplistic use case due to tech debt we have and how we process events of different "types".

PR #327, opened 2019-01-18
Additional Testing Around Blacklist and ConfigurationClient

electrostat:
Add additional testing (unit and instrumentation) for Blacklist and ConfigurationClient. Goal is to increase test coverage, handle corner cases better, and prepare for wider array of testing platforms (api levels + devices).

Fixes: https://github.com/mapbox/mapbox-events-android/issues/330

### DIFF
--- a/libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/ConfigurationClientInstrumentationTest.java
+++ b/libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/ConfigurationClientInstrumentationTest.java
@@ -1,0 +1,57 @@
+package com.mapbox.android.telemetry;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.support.test.InstrumentationRegistry;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.IOException;
+
+import okhttp3.Call;
+import okhttp3.OkHttpClient;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+
+public class ConfigurationClientInstrumentationTest {
+  private ConfigurationClient configurationClient;
+  private static final long DAY_IN_MILLIS = 86400000;
+
+  @Before
+  public void setup() {
+    Context context = InstrumentationRegistry.getTargetContext();
+    this.configurationClient = new ConfigurationClient(context,
+      TelemetryUtils.createFullUserAgent("AnUserAgent", context), "anAccessToken", new OkHttpClient());
+  }
+
+  @Test
+  public void shouldUpdateTest() {
+    setTimeStamp(System.currentTimeMillis() - DAY_IN_MILLIS);
+    assertTrue(configurationClient.shouldUpdate());
+
+    setTimeStamp(System.currentTimeMillis());
+    assertFalse(configurationClient.shouldUpdate());
+  }
+
+  @Test
+  public void saveTimeStampTest() {
+    setTimeStamp(System.currentTimeMillis() - DAY_IN_MILLIS);
+    Call mockedCall = mock(Call.class);
+    IOException mockedException = mock(IOException.class);
+
+    configurationClient.onFailure(mockedCall, mockedException);
+
+    assertFalse(configurationClient.shouldUpdate());
+  }
+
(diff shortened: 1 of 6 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
```diff
@@ -100,7 +101,9 @@ private static boolean isValidContent(String data) {
     JsonArray jsonArray;
     try {
       JsonObject responseJson = gson.fromJson(data, JsonObject.class);
-      jsonArray = responseJson.getAsJsonArray("RevokedCertKeys");
+      JsonElement jsonElement = responseJson.get("RevokedCertKeys");
```
andrlee: does `responseJson.getAsJsonArray("RevokedCertKeys");` returns `null` if object is not array? so why do we need additional helper on line 140?

libtelemetry/src/test/java/com/mapbox/android/telemetry/ConfigurationClientTest.java
```diff
@@ -26,6 +32,8 @@
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 public class ConfigurationClientTest {
```
andrlee: should we also test for a `null` handler passed into the client?

libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/ConfigurationClientInstrumentationTest.java
```diff
@@ -0,0 +1,57 @@
...
+    setTimeStamp(System.currentTimeMillis() - DAY_IN_MILLIS);
+    Call mockedCall = mock(Call.class);
+    IOException mockedException = mock(IOException.class);
+
+    configurationClient.onFailure(mockedCall, mockedException);
+
+    assertFalse(configurationClient.shouldUpdate());
+  }
+
+  private void setTimeStamp(Long milliseconds) {
```
andrlee: boxed type?

libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/ConfigurationClientInstrumentationTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.io.IOException;
+
+import okhttp3.Call;
+import okhttp3.OkHttpClient;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+
+public class ConfigurationClientInstrumentationTest {
```
andrlee: are we actually hitting real backend with this instrumentation test?

libtelemetry/src/main/java/com/mapbox/android/telemetry/CertificateBlacklist.java
electrostat: During testing, this did not hold true. Test failed when writing check with `IllegalStateException`. Wrote boolean check to correct that issue. Just dug into gson documentation to confirm and that is what it says. It did enlighten me to a method that already existed for this functionality though, so will replace with new method.

https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.5

libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/ConfigurationClientInstrumentationTest.java
electrostat: We are currently not hitting real backend. Spent 2+ hours working on solution and decided to get the ball rolling on the review process. Did not find solution to this amongst coworkers and online resources. `CountdownLatch`, which I used previously for similar Async tests is also not working in this situation. 

Please advise or point in right direction for hunting down a good solution on this.

andrlee: would this be helpful: https://developer.android.com/reference/androidx/test/espresso/idling/net/UriIdlingResource ?

PR #329, opened 2019-01-22
Define interface for data repository.

Chaoba:
@andrlee @electrostat We could discuss the design for data repository here. Please comment if you have any idea.


alfwatt:
Stale

PR #332, opened 2019-01-24
Added generic PerformanceEvent

osana:
added generic event to measure map performance

`data` could be `Bundle `or `Map<String, Object>`


langsmith:
Closing because @osana and I are going to be adding similar changes in the Maps SDK rather than in this repo.

PR #333, opened 2019-01-30
Release Telemetry SDK Lite on Maven

electrostat:
Addresses: https://github.com/mapbox/mobile-telemetry/issues/272

Change pipeline system to release `Lite` and `Full` SDK on maven. Currently neither is working properly due to `Lite` refactor.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,7 @@ jobs:
       IS_LOCAL_DEVELOPMENT: false
       OK_TO_RELEASE_CORE: true
       OK_TO_RELEASE_TELEM: true
+      OK_TO_RELEASE_LITE: false
     steps:
       - checkout
       - run:
@@ -117,6 +118,9 @@ jobs:
               fi
               if [ "$OK_TO_RELEASE_TELEM" == true ]; then
                 make publish-telem ;
+              fi
+              if [ "$OK_TO_RELEASE_LITE" == true ]; then
+                make publish-lite ;
               fi
             fi
       - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,7 +111,7 @@ jobs:
       - deploy:
           name: Publish to Maven Central
           command: |
-            if [ "${CIRCLE_BRANCH}" == "master" ]; then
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -93,6 +93,7 @@ jobs:
       IS_LOCAL_DEVELOPMENT: false
       OK_TO_RELEASE_CORE: true
       OK_TO_RELEASE_TELEM: true
+      OK_TO_RELEASE_LITE: false
```
andrlee: why not just release both at the same time? any potential issues you see?

Makefile
```diff
@@ -20,6 +20,9 @@ publish-core:
 publish-telem:
 	export IS_LOCAL_DEVELOPMENT=false; ./gradlew :libtelemetry:uploadArchives
 
+publish-lite:
+    export IS_LOCAL_DEVELOPMENT=false; ./gradlew :libtelemetry:uploadArchives
```
andrlee: does it look like it might be a dup of the above?

.circleci/config.yml
electrostat: work is still in progress. Working on being able to release separately and then will be easy enough to combine if necessary.

Makefile
electrostat: see above

gradle/mvn-push-android.gradle
```diff
@@ -97,6 +118,60 @@ afterEvaluate { project ->
...
+                        if (!VERSION_NAME.contains('LITE')) {
+                            String coreString = "com.mapbox.mapboxsdk:mapbox-android-core:" + core.version
+                            def coreArray = coreString.split(':')
+                            dependenciesArray.add(coreArray)
+
+                            String lifecycleString = dependenciesList.archLifecycleExtensions
+                            def lifecycleArray = lifecycleString.split(':')
+                            dependenciesArray.add(lifecycleArray)
+
+                            def viewModel = ["viewmodel", "android.arch.lifecycle"]
```
andrlee: why are we hardcoding all these? are we expected to hardcode every time we add/remove dependencies?

PR #335, opened 2019-01-31
Mapbox unhandled exception handler

andrlee:
The attempt to capture crash events caused by `libtelemetry` and `libcore` and report them to events endpoint. This metric is important to track for Telemetry SDK SLA. Note: this crash reporter is not capturing full stack traces generated by apps and is only designed to capture Mapbox java exceptions and metadata in privacy friendly way.

**Fatal Errors**
We're protecting ourselves from false positives by selecting the appropriate exception chain depth. `libtelemetry` for instance is usually built with maps or nav sdk, hence we're not looking at stack traces higher than on the 3rd level of the causal chain. 
```
HighLevelException: MidLevelException: LowLevelException
         at Junk.a(Junk.java:13) <- top of the stack, which is the last method invocation in the sequence
         at Junk.main(Junk.java:4)
 Caused by: MidLevelException: LowLevelException
         at Junk.c(Junk.java:23)
         at Junk.b(Junk.java:17)
         at Junk.a(Junk.java:11)
         ... 1 more
 Caused by: LowLevelException
         at Junk.e(Junk.java:30)
         at Junk.d(Junk.java:27)
         at Junk.c(Junk.java:21)
         ... 3 more
```
What about exceptions thrown in the context of system callback? Aren't they high level?
**Silent Errors**
System callbacks called by `JobService`, `BroadcastReceiver`, `Service`, `AlarmManager` and etc where mapbox code may fail will be enclosed by try and catch blocks and logged as a silent errors.
```
public void onSystemCallback() {
  try {
    //
  } catch (Throwable t) {
    MapboxCrashreporter.log(t);
  }
}
```
All these assumptions will be empirically validated with staging environment in the next few weeks.


libcore/src/main/java/com/mapbox/android/core/crashreporter/MapboxUncaughtExceptionHanlder.java
```diff
@@ -0,0 +1,124 @@
...
+      final StackTraceElement[] stackTraceElements = cause.getStackTrace();
+      for (final StackTraceElement e : stackTraceElements) {
+        String className = e.getClassName();
+        if (className.startsWith(mapboxPackage)) {
+          return true;
+        }
+      }
+      cause = cause.getCause();
+    }
+    return false;
```
tobrun: This will capture any crash that has some kind of mapbox package in the stacktrace. This however can also result in capturing not Mapbox related crashes. For example, a user makes a null pointer exception inside the OnMapReady callback with code unrelated to our SDKs. Afaik with above implementation this is flagged as a mapbox crash.

andrlee: @tobrun fair concern! this is a naive implementation, which i plan to generalize/improve as i dive deeper into extensive integration testing. based on my preliminary testing and given that telemetry sdk literally exposes 3 public apis (which i hope we can reduce to 1 in future), it sort of works :)

andrlee:
@alfwatt feel free to do post-commit review when you have time ;)

PR #336, opened 2019-02-01
Add more test cases for ConfigurationClient

Chaoba:
Fix one wrong test case, add two test cases to verify getting the correct endpoint


libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/ConfigurationClientInstrumentationTest.java
```diff
@@ -33,6 +42,44 @@ public void setup() {
       TelemetryUtils.createFullUserAgent("AnUserAgent", context), "anAccessToken", new OkHttpClient());
   }
 
+  @Test
+  public void checkComEndpoint() {
```
andrlee: what about `events.mapbox.com` and `events.mapbox.cn`?

Chaoba:
> Can you share mode details on what was wrong in the existing test?

In `updateRequestTest()`, the asset in `ConfigurationChangeHandler` will never be executed before the end the this test case.

Issue #312, opened 2018-12-07
LocationEngine doesn't respect fastest interval setting


LukasPaczos:
I'm seeing, that even after setting the `setFastestInterval` option the location updates can be pushed with higher than expected frequency.

Here's a log dump of differences in time between location samples with the fastest interval set to 750ms.
```
D/TEST: 1020
D/TEST: 15
D/TEST: 992
D/TEST: 1010
D/TEST: 1001
D/TEST: 972
D/TEST: 1016
D/TEST: 988
D/TEST: 12
D/TEST: 1012
D/TEST: 996
D/TEST: 1007
D/TEST: 1001
D/TEST: 1000
D/TEST: 983
D/TEST: 17
D/TEST: 1001
D/TEST: 998
D/TEST: 1005
D/TEST: 994
D/TEST: 989
D/TEST: 15
D/TEST: 1004
D/TEST: 991
D/TEST: 1009
D/TEST: 999
```

and here's the setup:
```
LocationEngine locationEngine = LocationEngineProvider.getBestLocationEngine(this);
locationEngine.requestLocationUpdates(
  new LocationEngineRequest.Builder(750)
    .setFastestInterval(750)
    .setPriority(LocationEngineRequest.PRIORITY_HIGH_ACCURACY)
    .build(),
  new LocationEngineCallback<LocationEngineResult>() {

    private long last = 0;

    @Override
    public void onSuccess(LocationEngineResult result) {
      long current = System.currentTimeMillis();
      long diff = current - last;
      last = current;
      Logger.d("TEST", String.valueOf(diff));
    }

    @Override
    public void onFailure(@NonNull Exception exception) {
    }
  },
  Looper.getMainLooper());
```

This particualr example was using `MapboxFusedLocationEngineImpl` and libtelemetry `v4.1.1`.

Chaoba:
By doing some tests, my preliminary conclusion is that
1. If we use `GoogleLocationEngineImpl`, the fastest interval  will works.
2. If we use `MapboxFusedLocationEngineImpl`, `setFastestInterval` will not works since it will pass the `Interval` parameter to location API and in framework it will set fastest interval with the same value of interval. 
3.In this case we set interval to 750ms and still could get location update in last than 750ms, since we pass the value correctly to System API, so I guess it is some thing wrong with framework, not our location sdk.

andrlee:
[...]

PR #338, opened 2019-02-12
Run liblocation test in CI

Chaoba:
Currently `liblocation` android instrumentation tests are not running as part of pre-commit tests, this PR will add them to CI and run on firebase.
What's more, rewrite test cases with mock.

### DIFF
--- a/liblocation/build.gradle
+++ b/liblocation/build.gradle
@@ -35,6 +35,7 @@ dependencies {
     testImplementation dependenciesList.assertJ
     testImplementation dependenciesList.gmsLocation
 
+    androidTestImplementation dependenciesList.mockitoAndroid
     androidTestImplementation(dependenciesList.testEspressoCore, {
         exclude group: 'com.android.support', module: 'support-annotations'
     })
--- a/liblocation/build.gradle
+++ b/liblocation/build.gradle
@@ -35,7 +35,6 @@ dependencies {
     testImplementation dependenciesList.assertJ
     testImplementation dependenciesList.gmsLocation
 
-    androidTestImplementation dependenciesList.mockitoAndroid
     androidTestImplementation(dependenciesList.testEspressoCore, {
         exclude group: 'com.android.support', module: 'support-annotations'
     })
--- a/liblocation/src/androidTest/java/com/mapbox/android/core/location/FusedLocationEngineInstrumentedTest.java
+++ b/liblocation/src/androidTest/java/com/mapbox/android/core/location/FusedLocationEngineInstrumentedTest.java
@@ -1,0 +1,130 @@
+package com.mapbox.android.core.location;
+
+import android.content.Context;
+import android.location.Criteria;
+import android.location.Location;
+import android.location.LocationListener;
+import android.location.LocationManager;
+import android.location.LocationProvider;
+import android.os.Looper;
+import android.support.annotation.NonNull;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.ArgumentMatchers.anyFloat;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(AndroidJUnit4.class)
+public class FusedLocationEngineInstrumentedTest {
+  private static final long INTERVAL = 1000L;
+  private static final String PROVIDER = "test_provider";
+  private ArrayList<LocationEngineProxy> engines = new ArrayList<>();
+  private LocationManager mockLocationManager;
+  private Location location = new Location(PROVIDER);
+
+  @Before
+  public void setUp() {
+    location.setLatitude(1.0);
+    location.setAltitude(2.0);
+    Context mockContext = mock(Context.class);
+    mockLocationManager = mock(LocationManager.class);
+    when(mockContext.getSystemService(anyString())).thenReturn(mockLocationManager);
+    List<String> providers = new ArrayList<>();
+    providers.add(PROVIDER);
+    when(mockLocationManager.getAllProviders()).thenReturn(providers);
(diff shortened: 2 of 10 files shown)


liblocation/src/androidTest/java/com/mapbox/android/core/location/FusedLocationEngineInstrumentedTest.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.ArgumentMatchers.anyFloat;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(AndroidJUnit4.class)
+public class FusedLocationEngineInstrumentedTest {
```
andrlee: shouldn't we try not to mock provider and location data in integration tests?

liblocation/src/androidTest/java/com/mapbox/android/core/location/FusedLocationEngineInstrumentedTest.java
```diff
@@ -0,0 +1,130 @@
...
+    for (LocationEngineProxy engineProxy : engines) {
+      engineProxy.requestLocationUpdates(getRequest(INTERVAL, LocationEngineRequest.PRIORITY_HIGH_ACCURACY),
+        engineCallback, null);
+
+      assertNotNull(engineProxy.removeListener(engineCallback));
+    }
+
+  }
+
+  private static LocationEngineRequest getRequest(long interval, int priority) {
```
andrlee: can we cover intervals, fastest intervals, max wait time with integration tests?

liblocation/src/androidTest/java/com/mapbox/android/core/location/GoogleLocationEngineInstrumentedTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(AndroidJUnit4.class)
+public class GoogleLocationEngineInstrumentedTest {
```
andrlee: same here we should not be mocking fused provider and perhaps start supporting https://developers.google.com/android/reference/com/google/android/gms/location/FusedLocationProviderClient.html#setMockLocation(android.location.Location) for testing purposes.

liblocation/src/androidTest/java/com/mapbox/android/core/location/FusedLocationEngineInstrumentedTest.java
Chaoba: It seems IdlingResources only works with a UI, but our testcases don't have UI, so it not works. Please correct me if I am wrong.

liblocation/src/androidTest/java/com/mapbox/android/core/location/FusedLocationEngineInstrumentedTest.java
Chaoba: These params are handled by android framework, so IMHO we don't need to test these.

andrlee: @Chaoba the location engines that we support are not in complete feature parity, hence not all configs are handled by framework. a good example is fastest interval: `MapboxFusedLocationEngine` doesn't support it: https://github.com/mapbox/mapbox-events-android/issues/312. Would be nice if we could create a test to cover all basic configs for future reference just in case someone decides to implement a new engine. Does that make sense?

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngineImpl.java
```diff
@@ -46,7 +45,8 @@ public void getLastLocation(@NonNull LocationEngineCallback<LocationEngineResult
     throws SecurityException {
     GoogleLastLocationEngineCallbackTransport transport =
       new GoogleLastLocationEngineCallbackTransport(callback);
-    fusedLocationProviderClient.getLastLocation().addOnSuccessListener(transport).addOnFailureListener(transport);
+    fusedLocationProviderClient.getLastLocation().addOnSuccessListener(transport);
+    fusedLocationProviderClient.getLastLocation().addOnFailureListener(transport);
```
andrlee: this would fetch last location twice, this is not what we want.

liblocation/src/main/java/com/mapbox/android/core/location/GoogleLocationEngineImpl.java
```diff
@@ -46,7 +46,9 @@ public void getLastLocation(@NonNull LocationEngineCallback<LocationEngineResult
     throws SecurityException {
     GoogleLastLocationEngineCallbackTransport transport =
       new GoogleLastLocationEngineCallbackTransport(callback);
-    fusedLocationProviderClient.getLastLocation().addOnSuccessListener(transport).addOnFailureListener(transport);
+    Task<Location> lastLocation = fusedLocationProviderClient.getLastLocation();
```
andrlee: i'm still a bit confused about this change, what difference does it make? the first call into `addOnSuccessListener(transport)` passes through instance of `Task<Location>` making this chain call possible.

liblocation/src/test/java/com/mapbox/android/core/location/GoogleLocationEngineImplTest2.java
```diff
@@ -0,0 +1,129 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+
```
andrlee: redundant line break

liblocation/src/test/java/com/mapbox/android/core/location/GoogleLocationEngineImplTest2.java
```diff
@@ -0,0 +1,129 @@
...
+public class GoogleLocationEngineImplTest2 {
```
andrlee: i wonder id we could come up with better name for this test class?

liblocation/build.gradle
```diff
@@ -35,6 +35,7 @@ dependencies {
     testImplementation dependenciesList.assertJ
     testImplementation dependenciesList.gmsLocation
 
+    androidTestImplementation dependenciesList.mockitoAndroid
```
andrlee: do we actually need this dependency?

Chaoba:
@andrlee All fixed now.

PR #341, opened 2019-02-15
Upload Module

electrostat:
Create module that is customizable for uploading collected telemetry data.


libupload/gradle.properties
```diff
@@ -0,0 +1,5 @@
+VERSION_NAME=0.1.0-SNAPSHOT
```
andrlee: i don't think we necessarily need to release upload library, i also think we could probably be just fine with implementing it as part of the internal package inside `libtelemetry`

libupload/src/main/java/com/mapbox/libupload/MapboxUploader.java
```diff
@@ -0,0 +1,28 @@
+package com.mapbox.libupload;
+
+import android.content.Context;
+
+public interface MapboxUploader <T, E> {
+
+  public void send(T data);
+
+  public void setConfiguration(Configuration configuration);
```
andrlee: method visibility modifier is redundant here

libtelemetry/src/full/java/com/mapbox/android/telemetry/UploadJobIntentService.java
```diff
@@ -0,0 +1,52 @@
...
+public class UploadJobIntentService extends JobIntentService implements Callback {
+  private static final String LOG_TAG = "UploadJobIntentService";
+  private static final int JOB_ID = 564321;
+  private static MapboxUploader.MapboxUploadClient<List<Event>, UploadJobIntentService> uploadClient;
+  private Gson gson = new Gson();
+
+  public void schedule(Context context, List<Event> events, MapboxUploader.MapboxUploadClient mapboxUploadClient) {
+    uploadClient = mapboxUploadClient;
+
+    Intent intent = new Intent("UploadJob");
```
andrlee: i wonder what's the reason behind fetching events prior to scheduling its upload?

libtelemetry/src/full/java/com/mapbox/android/telemetry/UploadClientFactory.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Map;
+
+class UploadClientFactory {
+  private static final String LOG_TAG = "TelemetryClientFactory";
+  private static final String RETRIEVING_APP_META_DATA_ERROR_MESSAGE = "Failed when retrieving app meta-data: %s";
+  private final String accessToken;
+  private final String userAgent;
+  private final CertificateBlacklist certificateBlacklist;
+  private final Context context;
+  private final Map<Environment, UploadClientBuild> BUILD_TELEMETRY_CLIENT = new HashMap<Environment,
```
andrlee: what's up with this map pattern again? :) can you please explain the benefit of it vs. simple `switch` in the static factory method?

PR #343, opened 2019-02-21
Aggregate instrumentation and unit test coverage

Chaoba:
This PR aggregates instrumentation test coverage with unit test coverage both locally via `build.sh` and `test.sh` and on CI via Firebase Test cloud.
The work is taking place due to the fact that our codecov.io coverage number doesn't include the instrumentation test coverage and we'd like to focus on improving our instrumentation test coverage in the short term.
Tested via `build.sh` and `test.sh` which we essentially run before pushing local changes to remote and testing in progress with Firebase test cloud.


cloud_test.sh
```diff
@@ -1,14 +1,19 @@
...
+    --device model=hammerhead,version=23,locale=en,orientation=landscape \
     --device model=sailfish,version=26,locale=en,orientation=portrait \
     --device model=sailfish,version=28,locale=en,orientation=portrait \
-    --timeout 20m
\ No newline at end of file
+    --environment-variables coverage=true,coverageFile="/sdcard/coverage.ec" \
+    --directories-to-pull /sdcard \
+    --timeout 20m
+
+covfile_path=`gsutil ls gs://test-lab-r47d1tyt8h0hm-iku3c1i8kjrux | tail -1`
```
Chaoba: Use `tail -1` has a potential  bug. If there are other tasks waiting in the firebase job queue, it will start a new job immediately after current job and `tail -1` will get this new job path.

andrlee: yeah that make sense. we'd probably need to create a permanent folder in the bucket for all coverage files.

PR #345, opened 2019-02-22
Add null check for pending Intent

Chaoba:
Run into a crash while testing

```
java.lang.NullPointerException: cancel() called with a null PendingIntent
at android.app.AlarmManager.cancel(AlarmManager.java:916)
at com.mapbox.android.telemetry.AlarmSchedulerFlusher.unregister(AlarmSchedulerFlusher.java:56)
at com.mapbox.android.telemetry.MapboxTelemetry.stopAlarm(MapboxTelemetry.java:433)
at com.mapbox.android.telemetry.MapboxTelemetry.stopTelemetry(MapboxTelemetry.java:538)
at com.mapbox.android.telemetry.MapboxTelemetry.disable(MapboxTelemetry.java:139)
at com.mapbox.android.events.testapp.MainActivity.onDestroy(MainActivity.java:37)
```
Root cause is that PendingIntent is inited in

```
 public void register() {
    Intent alarmIntent = receiver.supplyIntent();
    pendingIntent = PendingIntent.getBroadcast(context, 0, alarmIntent, 0);
    IntentFilter filter = new IntentFilter(SCHEDULER_FLUSHER_INTENT);
    context.registerReceiver(receiver, filter);
  }
```
If this method is not called before destroy, the pendingIntent will be null.

### DIFF
--- a/app/src/androidTest/java/com/mapbox/android/events/testapp/MainActivityInstrumentationTests.java
+++ b/app/src/androidTest/java/com/mapbox/android/events/testapp/MainActivityInstrumentationTests.java
@@ -1,0 +1,23 @@
+package com.mapbox.android.events.testapp;
+
+import android.app.Activity;
+import android.support.test.rule.ActivityTestRule;
+import android.support.test.runner.AndroidJUnit4;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertNotNull;
+
+@RunWith(AndroidJUnit4.class)
+public class MainActivityInstrumentationTests {
+  @Rule
+  public ActivityTestRule rule = new ActivityTestRule(MainActivity.class, true, true);
+
+  @Test
+  public void checkActivity() {
+    Activity activity = rule.getActivity();
+    assertNotNull(activity);
+  }
+}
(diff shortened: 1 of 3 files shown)


libtelemetry/src/testFull/java/com/mapbox/android/telemetry/AlarmSchedulerFlusherTest.java
```diff
@@ -61,7 +61,7 @@ public void checksAlarmUnregistered() throws Exception {
 
     theAlarmSchedulerFlusher.unregister();
 
-    verify(mockedAlarmManager, times(1)).cancel(eq(theAlarmSchedulerFlusher.obtainPendingIntent()));
+    verify(mockedAlarmManager, times(0)).cancel(eq(theAlarmSchedulerFlusher.obtainPendingIntent()));
```
andrlee: you may want to use `never()` here instead of `times(0)`

app/src/androidTest/java/com/mapbox/android/events/testapp/MainActivityInstrumentationTests.java
```diff
@@ -0,0 +1,23 @@
...
+import android.support.test.runner.AndroidJUnit4;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertNotNull;
+
+@RunWith(AndroidJUnit4.class)
+public class MainActivityInstrumentationTests {
```
andrlee: so what's the point of this test here?

Chaoba: Just to run throw the whole process in MainActivity. BTW, this bug was found by this test case, we could add more test cases in future in possible.

andrlee:
@Chaoba is this still in works?

Chaoba:
No further works.

PR #346, opened 2019-02-23
Execute network requests on background thread

andrlee:
Moving thread pool executor to `MapboxTelemetry` and using thread pool for all "expensive" operations that may be executed on the main thread.This is a temporary workaround until data repository is introduced and uploader/scheduler are refactored.

### DIFF
--- a/libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
@@ -23,6 +23,11 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.SynchronousQueue;
 
 import okhttp3.Call;
 import okhttp3.Callback;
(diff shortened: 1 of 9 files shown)


libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -601,4 +617,23 @@ public void onResponse(Call call, Response response) throws IOException {
...
+      return new ThreadPoolExecutor(0, maxSize,
+        keepAliveSeconds, TimeUnit.SECONDS, new SynchronousQueue<Runnable>(),
+        threadFactory(name));
+    }
+
+    private static ThreadFactory threadFactory(final String name) {
+      return new ThreadFactory() {
+        @Override
+        public Thread newThread(Runnable runnable) {
+          return new Thread(runnable, name);
```
Chaoba: For every thread, the name will be the same, will be nice if we add a suffix.

andrlee: what would suffix be helpful for?

Chaoba: So we can identify each thread and will be helpful during our debug. The default construct method of Thread class will add a suffix like this
```
public Thread() {
       init(null, null, "Thread-" + nextThreadNum(), 0);
}
```

andrlee: Still a bit vague on how thread number will help in debugging specifically in the context of this thread pool, but for the record if one of these threads crash, we will capture `tid`, `name` and its `priority`. So let's move on with this PR unless you have any other feedback.

Chaoba: OK, let's move on.

PR #349, opened 2019-02-26
Location collector refactor

andrlee:
This is a first iteration of location collector refactor, which intends to replace `TelemetryService` and put in place a foundation for state machine that will eventually be responsible for complex state management of our location trace collection. Conceptually this change decouples location trace collection from reporting part of the telemetry stack and turns location collector into an internal customer of `MapboxTelemetry` similarly to how maps or nav use telemetry library.

/cc @mapbox/mobile-telemetry


libtelemetry/src/full/java/com/mapbox/android/telemetry/location/LocationUpdatesJobIntentService.java
```diff
@@ -0,0 +1,59 @@
...
+public class LocationUpdatesJobIntentService extends JobIntentService {
+  private static final int JOB_ID = 102;
+  private static final String TAG = "LocJobIntentService";
+
+  public static void enqueueWork(Context context, Intent intent) {
+    enqueueWork(context, LocationUpdatesJobIntentService.class, JOB_ID, intent);
+  }
+
+  @Override
+  protected void onHandleWork(@NonNull Intent intent) {
```
Chaoba: It seems that we don't have time consuming work here, so why not do these works in `LocationUpdatesBroadcastReceiver` directly?

libtelemetry/src/androidTestFull/java/com/mapbox/android/telemetry/location/LocationCollectionClientInstrumentedTest.java
```diff
@@ -0,0 +1,70 @@
...
+  private LocationCollectionClient ref;
+
+  @Before
+  public void setUp() {
+    ref = LocationCollectionClient.install(InstrumentationRegistry.getTargetContext(),
+      1000);
+  }
+
+  @After
+  public void tearDown() {
```
Chaoba: Since it is a Singleton, `ref = null` will do nothing to it. We should call `uninstall` in this method.

libtelemetry/src/full/java/com/mapbox/android/telemetry/location/LocationUpdatesBroadcastReceiver.java
```diff
@@ -0,0 +1,33 @@
...
+      if (intent == null) {
+        Log.w(TAG, "intent == null");
+        return;
+      }
+
+      final String action = intent.getAction();
+      if (ACTION_LOCATION_UPDATED.equals(action)) {
+        LocationUpdatesJobIntentService.enqueueWork(context, intent);
+      }
+    } catch (Throwable throwable) {
```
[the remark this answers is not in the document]
andrlee: this is a ground work for our in-house crash reporting...i couldn't find a consistent mechanism to infer high-level exceptions thrown by our code, so the workaround going forward will be to enclose all system callbacks and log crashes as silent.

libtelemetry/src/full/java/com/mapbox/android/telemetry/location/LocationUpdatesJobIntentService.java
andrlee: I agree it might be an overkill, will revisit in the next iteration, i did not write any unit tests for it for the reason that i'd be profiling it right after i integrate it with mapbox telem.

libtelemetry/src/androidTestFull/java/com/mapbox/android/telemetry/location/LocationCollectionClientInstrumentedTest.java
andrlee: yeah, good catch thanks!

PR #350, opened 2019-02-26
Fix GoogleLocationEngine intermittent delayed updates

andrlee:
Reset default `displacement` value to 0 due to potential delay with first location update on Pie that  takes 30 seconds - 1 min. We've not seen this behavior on pre-Pie api versions, so the assumption is that changes in framework were made that no longer work well with default displacement value we had tested with previously.

/cc @danesfeder @Guardiola31337

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -7,7 +7,7 @@ ext {
     ]
 
     version = [
-            gmsLocation         : '15.0.1',
+            gmsLocation         : '16.0.0',
             junit               : '4.12',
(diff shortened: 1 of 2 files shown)


andrlee:
@Guardiola31337 i was not able to repro this on emulator running Pie and i don't have hardware with Pie installed at hand. I asked @electrostat to help with testing though. May i suggest something? Can you ask your customers who have reported this to explicitly `setDisplacement` to `0` in their `LocationEngineRequest` object and retest?

PR #351, opened 2019-03-04
Sunset telemetry service

andrlee:
This refactor migrates location traces collection feature from being backed by Android `Service`.
Telemetry location traces collection relies on `Service` to do background work, this approach no longer works on apps that target Android 8.0 (API level 26) or higher due to OS limitations imposed on what apps can do while running in the background: https://developer.android.com/about/versions/oreo/background. 

**Refactor objectives:** 
- Support location data collection while app is in background on Oreo and above.
- Extract location collection component from core telemetry.
- Achieve dependency inversion between core telemetry and location components.
- Encapsulate location collection controller and build foundation for the state machine.
- Integrate with existing mapbox core telemetry in backwards compatible way.


libtelemetry/src/full/java/com/mapbox/android/telemetry/location/LocationCollectionClient.java
```diff
@@ -68,7 +84,10 @@ public static LocationCollectionClient install(@NonNull Context context, long de
       if (locationCollectionClient == null) {
         locationCollectionClient = new LocationCollectionClient(new LocationEngineControllerImpl(applicationContext,
           LocationEngineProvider.getBestLocationEngine(applicationContext), new SessionIdentifier(defaultInterval)),
-          new HandlerThread("LocationSettingsChangeThread"));
+          new HandlerThread("LocationSettingsChangeThread"),
+          applicationContext.getSharedPreferences(MAPBOX_SHARED_PREFERENCES, Context.MODE_PRIVATE),
+          // Provide empty token as it is not available yet
+          new MapboxTelemetry(applicationContext, "", LOCATION_COLLECTOR_USER_AGENT));
```
Chaoba: I am confused here as MapboxTelemetry is the API class we provide for MapSdk and AutoSDK, they will create a instance of MapboxTelemetry to push gesture events etc, but we create another instance here.

libtelemetry/src/full/java/com/mapbox/android/telemetry/provider/MapboxTelemetryInitProvider.java
```diff
@@ -0,0 +1,78 @@
...
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+import com.mapbox.android.telemetry.location.LocationCollectionClient;
+
+import java.util.concurrent.TimeUnit;
+
+import static com.mapbox.android.telemetry.location.LocationCollectionClient.DEFAULT_SESSION_ROTATION_INTERVAL_HOURS;
+
+public class MapboxTelemetryInitProvider extends ContentProvider {
```
Chaoba: Why do we need a ContentProvider here? Just for init?

andrlee: yes, it's a hacky way to execute your code at app startup

libtelemetry/src/full/java/com/mapbox/android/telemetry/location/LocationCollectionClient.java
andrlee: `MapboxTelemetry` is no longer coupled with location collection and hence becomes responsible for eventing only. You can treat `LocationCollectionClient` as producer of location events, which are delivered to the `api-events` endpoint via `MapboxTelemetry`.

libtelemetry/src/full/java/com/mapbox/android/telemetry/location/LocationCollectionClient.java
```diff
@@ -83,8 +84,9 @@ public static LocationCollectionClient install(@NonNull Context context, long de
     synchronized (lock) {
       if (locationCollectionClient == null) {
         locationCollectionClient = new LocationCollectionClient(new LocationEngineControllerImpl(applicationContext,
-          LocationEngineProvider.getBestLocationEngine(applicationContext), new SessionIdentifier(defaultInterval)),
+          LocationEngineProvider.getBestLocationEngine(applicationContext)),
           new HandlerThread("LocationSettingsChangeThread"),
+          new SessionIdentifier(defaultInterval),
           applicationContext.getSharedPreferences(MAPBOX_SHARED_PREFERENCES, Context.MODE_PRIVATE),
```
Chaoba: We should use `TelemetryUtils.obtainSharedPreferences` here

andrlee: It is intentional: `TelemetryUtils` should not be public, the plan is to downgrade its visibility to package private at some point. Hence i prefer to avoid creating any dependencies on it going forward.

Chaoba: Ok, but keep in mind we may have strict mode issue here.

andrlee: Are you suggesting to access it on the background thread?

Chaoba: Yeah, my thought is that we could create a sharedPreference later and put all background threads in it.

andrlee: i need shared preferences at startup but i will make sure there's no impact on cold start by deferring init to background thread.

PR #352, opened 2019-03-04
Implement the function of setFastestInterval

### LINKED ISSUE #312: LocationEngine doesn't respect fastest interval setting

Chaoba:
Resolves #312 
`LocationManager` will not receive `fastestInterval` parameter, so currently we only pass `interval` value to it. What's more `LocationManager` will not respect this parameter, even we set it to 1000ms, we could still receive update in less than 1000ms.
According [android doc](https://developer.android.com/training/location/change-location-settings#location-request), we shouldn't receive location update if set `fastestInterval`.
This PR will fix this issue by adding downsample in `LocationEngineImpl`

### DIFF
--- a/app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
+++ b/app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
@@ -1,21 +1,20 @@
 package com.mapbox.android.events.testapp;
 
 import android.annotation.SuppressLint;
-import android.os.Bundle;
-import android.os.Looper;
 import android.location.Location;
 import android.support.annotation.NonNull;
 import android.support.v7.app.AppCompatActivity;
-import android.util.Log;
+import android.os.Bundle;
 
+import android.util.Log;
 import android.view.View;
 import android.widget.Button;
 import android.widget.Toast;
 import com.mapbox.android.core.location.LocationEngine;
 import com.mapbox.android.core.location.LocationEngineCallback;
+import com.mapbox.android.core.location.LocationEngineResult;
 import com.mapbox.android.core.location.LocationEngineProvider;
 import com.mapbox.android.core.location.LocationEngineRequest;
-import com.mapbox.android.core.location.LocationEngineResult;
 import com.mapbox.android.core.permissions.PermissionsListener;
 import com.mapbox.android.core.permissions.PermissionsManager;
 import com.mapbox.android.telemetry.AppUserTurnstile;
@@ -98,31 +97,6 @@ public class MainActivity extends AppCompatActivity implements PermissionsListen
     boolean permissionsGranted = PermissionsManager.areLocationPermissionsGranted(this);
     if (permissionsGranted) {
       mapboxTelemetry.enable();
-      LocationEngine locationEngine = LocationEngineProvider.getBestLocationEngine(this);
-      locationEngine.requestLocationUpdates(
-        new LocationEngineRequest.Builder(750)
-          .setFastestInterval(750)
-          .setPriority(LocationEngineRequest.PRIORITY_HIGH_ACCURACY)
-          .build(),
-        new LocationEngineCallback<LocationEngineResult>() {
-
-          private long last = 0;
-
-          @Override
-          public void onSuccess(LocationEngineResult result) {
-            long current = System.currentTimeMillis();
-            long diff = current - last;
-            last = current;
-            String msg = String.valueOf(diff);
-            Log.d("TEST", msg);
-          }
-
-          @Override
-          public void onFailure(@NonNull Exception exception) {
-          }
-        },
-        Looper.getMainLooper());
-
     } else {
       PermissionsManager permissionsManager = new PermissionsManager(this);
       permissionsManager.requestLocationPermissions(this);
--- a/app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
+++ b/app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
@@ -7,7 +7,6 @@ import android.location.Location;
 import android.support.annotation.NonNull;
 import android.support.v7.app.AppCompatActivity;
 import android.util.Log;
-import android.widget.Toast;
 
 import android.view.View;
 import android.widget.Button;
--- a/app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
+++ b/app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
@@ -1,20 +1,22 @@
 package com.mapbox.android.events.testapp;
 
 import android.annotation.SuppressLint;
+import android.os.Bundle;
+import android.os.Looper;
 import android.location.Location;
 import android.support.annotation.NonNull;
 import android.support.v7.app.AppCompatActivity;
-import android.os.Bundle;
-
 import android.util.Log;
+import android.widget.Toast;
+
 import android.view.View;
 import android.widget.Button;
 import android.widget.Toast;
 import com.mapbox.android.core.location.LocationEngine;
 import com.mapbox.android.core.location.LocationEngineCallback;
-import com.mapbox.android.core.location.LocationEngineResult;
 import com.mapbox.android.core.location.LocationEngineProvider;
 import com.mapbox.android.core.location.LocationEngineRequest;
+import com.mapbox.android.core.location.LocationEngineResult;
 import com.mapbox.android.core.permissions.PermissionsListener;
 import com.mapbox.android.core.permissions.PermissionsManager;
 import com.mapbox.android.telemetry.AppUserTurnstile;
@@ -97,6 +99,31 @@ public class MainActivity extends AppCompatActivity implements PermissionsListen
     boolean permissionsGranted = PermissionsManager.areLocationPermissionsGranted(this);
     if (permissionsGranted) {
       mapboxTelemetry.enable();
+      LocationEngine locationEngine = LocationEngineProvider.getBestLocationEngine(this);
+      locationEngine.requestLocationUpdates(
+        new LocationEngineRequest.Builder(750)
+          .setFastestInterval(750)
+          .setPriority(LocationEngineRequest.PRIORITY_HIGH_ACCURACY)
+          .build(),
+        new LocationEngineCallback<LocationEngineResult>() {
+
+          private long last = 0;
+
+          @Override
+          public void onSuccess(LocationEngineResult result) {
+            long current = System.currentTimeMillis();
+            long diff = current - last;
+            last = current;
+            String msg = String.valueOf(diff);
+            Log.d("TEST", msg);
+          }
+
+          @Override
+          public void onFailure(@NonNull Exception exception) {
+          }
+        },
+        Looper.getMainLooper());
+
     } else {
       PermissionsManager permissionsManager = new PermissionsManager(this);
       permissionsManager.requestLocationPermissions(this);
--- a/liblocation/src/main/java/com/mapbox/android/core/location/AndroidLocationEngineImpl.java
+++ b/liblocation/src/main/java/com/mapbox/android/core/location/AndroidLocationEngineImpl.java
@@ -67,6 +67,7 @@ class AndroidLocationEngineImpl implements LocationEngineImpl<LocationListener> 
                                      @Nullable Looper looper) throws SecurityException {
     // Pick best provider only if user has not explicitly chosen passive mode
     currentProvider = getBestProvider(request.getPriority());
+    ((LocationCallbackTransport) listener).setFastInterval(request.getFastestInterval());
     locationManager.requestLocationUpdates(currentProvider, request.getInterval(), request.getDisplacemnt(),
       listener, looper);
   }
@@ -83,6 +84,7 @@ class AndroidLocationEngineImpl implements LocationEngineImpl<LocationListener> 
   @Override
   public void removeLocationUpdates(@NonNull LocationListener listener) {
     if (listener != null) {
+      ((LocationCallbackTransport) listener).onDestroy();
       locationManager.removeUpdates(listener);
     }
   }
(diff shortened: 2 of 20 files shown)


liblocation/src/main/java/com/mapbox/android/core/location/AndroidLocationEngineImpl.java
```diff
@@ -141,13 +142,23 @@ private static int priorityToPowerRequirement(int priority) {
...
-    AndroidLocationEngineCallbackTransport(LocationEngineCallback<LocationEngineResult> callback) {
+    AndroidLocationEngineCallbackTransport(@NonNull LocationEngineRequest request,
+                                           LocationEngineCallback<LocationEngineResult> callback) {
+      this.request = request;
       this.callback = callback;
     }
 
     @Override
     public void onLocationChanged(Location location) {
+      long currentTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
Chaoba: In testcase it would always be 0 and fail our test, so let's keep it.

app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
```diff
@@ -70,7 +72,32 @@ private void checkPermissions() {
 
     if (permissionsGranted) {
       mapboxTelemetry.enable();
-      locationEngine.requestLocationUpdates(getRequest(2000), callback, null);
+      LocationEngine locationEngine = LocationEngineProvider.getBestLocationEngine(this);
```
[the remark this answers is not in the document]
Chaoba: locationEngine is a temporary variable, so it will be released out of this block and will not result in memory leak.

LukasPaczos:
This solution is similar to what we've implemented in the `LocationComponent` as a workaround and we've not heard any complaints for the past 2 months or so. However, what I dislike about the precision of this approach is that we can often run into a situation where the update is going to be skipped completely because it's missed the window by a couple of milliseconds, leaving end users with a gap that's double the interval length.

What do you think about introducing some kind of a buffer that would start to cache location samples provided in the last `x`% of the interval length and deliver the latest (if any) of those right when we cross the threshold? This would provide a better sense of continuity, even when the location samples coming from the framework are not precisely matching the interval.

liblocation/src/main/java/com/mapbox/android/core/location/AndroidLocationEngineImpl.java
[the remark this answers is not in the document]
andrlee: What do you mean? should our tests drive internal implementation details? the test is failing because `SystemClock` is android framework class and we're using `unitTests.returnDefaultValues = true` by default. That said `System.currentTimeMillis()` is not safe for measuring time spans, so let's try to find workaround.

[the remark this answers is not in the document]
Chaoba: In Unit test `SystemClock.elapsedRealtime()` will always be 0, but `System.currentTimeMillis()` can get the real time.

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
```diff
@@ -13,30 +17,62 @@
...
   private Location currentBestLocation;
-  private long lastUpdateTime = 0;
   private long fastestInterval = 0;
+  private Location cacheLocation;
+  private Handler handler;
+  private HandlerThread handlerThread;
 
   LocationCallbackTransport(LocationEngineCallback<LocationEngineResult> callback) {
     this.callback = callback;
+    handlerThread = new HandlerThread(this.getClass().getName());
```
LukasPaczos: Why do we need a separate thread here? It forces the location update delivery on a background thread which we should avoid at all cost. It should be just fine to process the `handleMessage` logic on the main thread.

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
```diff
@@ -13,30 +17,62 @@
...
+      super(looper);
+    }
+
+    @Override
+    public void handleMessage(Message msg) {
+      Location location = cacheLocation;
+      cacheLocation = null;
+      if (location != null && sendLocation(location)) {
+        //send location success, start to send next one after fastestInterval period
+        this.sendEmptyMessageDelayed(MESSAGE_WHAT, fastestInterval);
```
LukasPaczos: This can be moved to the `sendLocation` method.

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
```diff
@@ -13,30 +17,62 @@
...
 
   @Override
   public void onLocationChanged(Location location) {
-    if (isBetterLocation(location, currentBestLocation)) {
-      currentBestLocation = location;
+    if (fastestInterval != 0) {
+      if (!handler.hasMessages(MESSAGE_WHAT)) {
+        //haven't send message in the previous fastInterval period, so send it directly.
+        sendLocation(location);
+        handler.sendEmptyMessageDelayed(MESSAGE_WHAT, fastestInterval);
```
LukasPaczos: Same as above.

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
```diff
@@ -13,30 +17,62 @@
...
+      } else {
+        //cacheLocation will always be the latest one.
+        cacheLocation = location;
```
LukasPaczos: We should cache only the location samples that come close to the interval finish time. Maybe a solution can be posting another message delayed by a fraction of the `fastestInterval` and cache location samples only if that message is not in the queue?

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
```diff
@@ -17,62 +15,63 @@
...
+          break;
+        default:
+          break;
       }
     }
   }
 
   @Override
   public void onLocationChanged(Location location) {
     if (fastestInterval != 0) {
```
LukasPaczos: There's no need to check for the `fastestInterval` here.

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
```diff
@@ -100,7 +99,6 @@ void setFastInterval(long fastestInterval) {
   }
 
   void onDestroy() {
-    handler.removeMessages(MESSAGE_WHAT);
-    handlerThread.quit();
+    handler.removeMessages(MESSAGE_UPDATE_LOCATION);
```
LukasPaczos: `MESSAGE_NOT_CACHE_PERIOD` should be removed as well to keep things clean.

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
```diff
@@ -0,0 +1,112 @@
...
+  private long fastestInterval = 0;
+  private Location cacheLocation;
+  private final Handler handler;
+
+  LocationCallbackTransport(LocationEngineCallback<LocationEngineResult> callback) {
+    this.callback = callback;
+    handler = new ThresholdHandler(this);
+  }
+
+  static class ThresholdHandler extends Handler {
```
andrlee: i wonder what are the benefits of having  this inner class non-private?

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
```diff
@@ -0,0 +1,112 @@
...
+    public void handleMessage(Message msg) {
+      LocationCallbackTransport callbackTransport = weakReference.get();
+      if (callbackTransport == null) {
+        return;
+      }
+      switch (msg.what) {
+        case MESSAGE_UPDATE_LOCATION:
+          Location location = callbackTransport.cacheLocation;
+          callbackTransport.cacheLocation = null;
+          callbackTransport.sendLocation(location);
```
andrlee: i wonder what are the benefits of dispatching messages to the main thread?

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
```diff
@@ -0,0 +1,112 @@
...
+      //haven't send message in the previous fastInterval period, so send it directly.
+      sendLocation(location);
+    } else {
+      if (!handler.hasMessages(MESSAGE_NOT_CACHE_PERIOD)) {
+        cacheLocation = location;
+      }
+    }
+  }
+
+  private void sendLocation(Location location) {
```
andrlee: we're violating api contract here by calling `sendLocation` on the main thread, consider the case when user provided its background thread `Looper` on which they want this callback to be dispatched on.

app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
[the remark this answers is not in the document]
LukasPaczos: It showcases a wrong setup which will result in leaks as @andrlee mentioned. Since we are aware of this, we should not include the example at all now and follow up in a separate PR.

liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProxy.java
```diff
@@ -31,7 +31,11 @@ public void requestLocationUpdates(@NonNull LocationEngineRequest request,
                                      @Nullable Looper looper) throws SecurityException {
     checkNotNull(request, "request == null");
     checkNotNull(callback, "callback == null");
-    locationEngineImpl.requestLocationUpdates(request, getListener(callback),
+    T listener = getListener(callback);
+    if (listener instanceof LocationCallbackTransport) {
```
andrlee: so, why'd we cast every `listener` of every location request when we'd need this only for `AndroidLocationEngine` and its children?

liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProxy.java
```diff
@@ -45,6 +49,12 @@ public void requestLocationUpdates(@NonNull LocationEngineRequest request,
   @Override
   public void removeLocationUpdates(@NonNull LocationEngineCallback<LocationEngineResult> callback) {
     checkNotNull(callback, "callback == null");
+    if (listeners != null) {
```
andrlee: same question as above...

app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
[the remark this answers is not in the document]
andrlee: @LukasPaczos the right setup with other goodies just landed in master: https://github.com/mapbox/mapbox-events-android/pull/347

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
Chaoba: Good catch, then we need to call this method on the same thread  that `onLocationChanged` is called.

liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProxy.java
Chaoba: I see, you mean we handle this logic in `AndroidLocationEngineImp`?

liblocation/src/main/java/com/mapbox/android/core/location/LocationCallbackTransport.java
```diff
@@ -0,0 +1,118 @@
...
+  LocationCallbackTransport(LocationEngineCallback<LocationEngineResult> callback) {
+    this.callback = callback;
+  }
+
+  private static class ThresholdHandler extends Handler {
+    private final WeakReference<LocationCallbackTransport> weakReference;
+
+    ThresholdHandler(Looper looper, LocationCallbackTransport callbackTransport) {
+      super(looper);
+      weakReference = new WeakReference<>(callbackTransport);
```
andrlee: under which conditions we're expecting that strong reference to callback may cause a leak?

liblocation/src/main/java/com/mapbox/android/core/location/AndroidLocationEngineImpl.java
```diff
@@ -68,6 +67,7 @@ public void requestLocationUpdates(@NonNull LocationEngineRequest request,
                                      @Nullable Looper looper) throws SecurityException {
     // Pick best provider only if user has not explicitly chosen passive mode
     currentProvider = getBestProvider(request.getPriority());
+    ((LocationCallbackTransport) listener).setFastInterval(request.getFastestInterval());
```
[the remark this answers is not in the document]
Chaoba: There is no case could listener be a type other than `LocationCallbackTransport`, so I think it is no need to handle it.

alfwatt:
Stale

PR #361, opened 2019-03-13
Crash reporter telemetry integration 

andrlee:
This PR integrates mapbox crash reporter into telemetry sdk. The purpose of this work is to provide insights into how reliable telemetry sdk is in production. 

Crash collection and reporting components live in different packages atm but they are designed with the modularity in mind, they both are extremely easy to extract into a library module if needed. At app startup `CrashReporterClient` is responsible for checking in crash events dumped on disk by `MapboxUncaughtExceptionHandler`. It does so in background via `JobInternService` which is triggered by broadcast sent upon valid token becomes available.

**Testing:**
Tested with devices pre and post Oreo. Ran tests with maps sdk test app. Also tested with staging api endpoint.

### DIFF
--- a/libtelemetry/src/full/java/com/mapbox/android/telemetry/crash/CrashReporterClient.java
+++ b/libtelemetry/src/full/java/com/mapbox/android/telemetry/crash/CrashReporterClient.java
@@ -54,7 +54,7 @@ final class CrashReporterClient {
 
   CrashReporterClient loadFrom(@NonNull File rootDir) {
     fileCursor = 0;
-    crashReports = listAllFiles(rootDir);
+    crashReports = FileUtils.listAllFiles(rootDir);
     Arrays.sort(crashReports, new FileUtils.LastModifiedComparator());
     return this;
   }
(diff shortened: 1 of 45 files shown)


libtelemetry/src/full/java/com/mapbox/android/telemetry/crash/CrashReporterClient.java
```diff
@@ -0,0 +1,165 @@
...
+
+  static CrashReporterClient create(@NonNull Context context) {
+    SharedPreferences sharedPreferences =
+      context.getSharedPreferences(MAPBOX_CRASH_REPORTER_PREFERENCES, Context.MODE_PRIVATE);
+    return new CrashReporterClient(context, sharedPreferences,
+      new MapboxTelemetry(context, "", CRASH_REPORTER_CLIENT_USER_AGENT), new File[0]);
+  }
+
+  CrashReporterClient load(@NonNull String rootPath) {
+    crashReports = FileUtils.listAllFiles(applicationContext, rootPath);
```
alfwatt: Should `fileCursor` be reset to `0` at this point?

andrlee: yeah fair point, i do initialize it to zero and expectation is that `load` is called once but if we'd reuse it to load other batch at the same time then it has to be reset to zero. did not have chance to write tests for it yet but thanks for the note!

alfwatt: I'd encapsulate it into a setter for the property

Issue #203, opened 2018-08-28
Remove map event creation code from the mapbox-events-android library


lilykaiser:
Per internal conversation, step 3 in the effort to streamline event ingestion is to pull Map Events creation code (classes, factories, test, etc...) out of the Telemetry Repo. (In parallel with [putting it in mapbox-gl-native](https://github.com/mapbox/mapbox-gl-native/issues/12757)). Let's do it!

Related nav ticket: https://github.com/mapbox/mapbox-events-android/issues/204

PR #373, opened 2019-03-29
Remove map event creation code

Chaoba:
Fix https://github.com/mapbox/mapbox-events-android/issues/203.
Remove map event creation code and corresponding test code.

The corresponding PR for map sdk is https://github.com/mapbox/mapbox-gl-native/pull/14309


andrlee:
hey @Chaoba thanks for starting looking into it. a quick note here: the gist of the ticket is to move all map events to `mapbox-gl-native` while maintaining backwards compatibility. so essentially we need to cut two PRs: one that removes all maps events from events repo and another one that adds them to maps sdk. Please consult with @tobrun or @LukasPaczos about what is the right place for these events

Chaoba:
@andrlee We should also create a PR for auto, right?

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapBuildEvent.java
```diff
@@ -1,6 +1,6 @@
 package com.mapbox.android.telemetry;
 
 
-interface MapBuildEvent {
+public interface MapBuildEvent {
```
andrlee: hey @Chaoba thanks for continuing looking into this. A quick note here: can we have all `Map*.java` classes moved to `gl-native` repo instead of exposing them publicly from telem sdk?

libtelemetry/src/main/java/com/mapbox/android/telemetry/Event.java
```diff
@@ -17,10 +17,11 @@
     VIS_GENERAL, VIS_ATTACHMENT, VIS_OBJ_DETECTION, NO_OP, CRASH
   }
 
-  static EnumSet<Type> mapGestureEventTypes = EnumSet.of(Type.MAP_CLICK, Type.MAP_DRAGEND);
-  static EnumSet<Type> navigationEventTypes = EnumSet.of(Type.NAV_DEPART, Type.NAV_ARRIVE, Type.NAV_CANCEL,
+  public static EnumSet<Type> mapGestureEventTypes = EnumSet.of(Type.MAP_CLICK, Type.MAP_DRAGEND);
```
andrlee: should these be moved altogether to `gl-native`?

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationFasterRouteEvent.java
```diff
@@ -5,7 +5,7 @@
 
 import com.google.gson.annotations.JsonAdapter;
 
-class NavigationFasterRouteEvent extends Event implements Parcelable {
+public class NavigationFasterRouteEvent extends Event implements Parcelable {
```
andrlee: i'd suggest to tackle all `Navigation*.java` events in a separate PR

libtelemetry/src/main/java/com/mapbox/android/telemetry/MapBuildEvent.java
Chaoba: That sounds good.

libtelemetry/src/main/java/com/mapbox/android/telemetry/NavigationFasterRouteEvent.java
Chaoba: Ok, will revert changes about Nav.

libtelemetry/src/main/java/com/mapbox/android/telemetry/Event.java
Chaoba: Will not use.

PR #377, opened 2019-04-01
Fix crash file counting bug

andrlee:
This fixes handling of cleanup logic when we reach max allowed reports in the root directory. 
Fix misused `File#length` api, with listing all files in the root directory and counting size of its array.
Tested with instrumentation tests and manually with maps sdk test app.

### DIFF
--- a/libcore/src/androidTest/java/com/mapbox/android/core/crashreporter/MapboxUncaughtExceptionHanlderInstrumentationTest.java
+++ b/libcore/src/androidTest/java/com/mapbox/android/core/crashreporter/MapboxUncaughtExceptionHanlderInstrumentationTest.java
@@ -51,18 +51,35 @@ public class MapboxUncaughtExceptionHanlderInstrumentationTest {
 
   @Test
   public void testReportCap() throws IOException {
-    for (int i = 0; i < 10; i++) {
-      File file = FileUtils.getFile(context,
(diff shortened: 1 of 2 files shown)


libcore/src/androidTest/java/com/mapbox/android/core/crashreporter/MapboxUncaughtExceptionHanlderInstrumentationTest.java
```diff
@@ -65,6 +60,28 @@ public void testReportCap() throws IOException {
...
+    writeToDisk(3);
+    MapboxUncaughtExceptionHanlder.ensureDirectoryWritable(context, TELEM_MAPBOX_PACKAGE);
+    assertEquals(3, directory.listFiles().length);
+  }
+
+  @Test
+  public void ensureDirectoryWritableCleansUp() throws IOException {
+    writeToDisk(10);
+    MapboxUncaughtExceptionHanlder.ensureDirectoryWritable(context, TELEM_MAPBOX_PACKAGE);
+    assertEquals(1, directory.listFiles().length);
```
Chaoba: Shell we check the left one is the latest one?

andrlee: good point - i'll add that test to the next PR.

PR #380, opened 2019-04-08
Add skuId to the turnstile events

andrlee:
Add `SKU` for to the turnstile event to prevent double billing. Also this change exposes a public setter `AppUserTurnstile#setSkuId(String)`, which should be used by our maps sdk to set a default value of `00` as per spec in order to address potential double billing. 

/cc @tobrun @zugaldia and @Guardiola31337 just in case it has impact on navigation sdk.


zugaldia:
Once this PR lands, and a release is out, this is the Maps SDK PR where we plan on integrating this new setter: https://github.com/mapbox/mapbox-gl-native/pull/14404.

PR #383, opened 2019-04-12
Remove nav event creation  related codes

### LINKED ISSUE #204: Remove nav event creation code from the mapbox-events-android library
Per internal conversation, step 4 in the effort to streamline tracking event ingestion is to refactor event creation code (classes, factories, test, etc...) [into the nav SDK](https://github.com/mapbox/mapbox-navigation-android/issues/1246) and out of mapbox-android-events. Creating this ticket as a placeholder.

Related maps ticket: https://github.com/mapbox/mapbox-events-android/issues/203

Chaoba:
Fix #204 
Remove nav event creation related code and corresponding test code.

Corresponding [PR](https://github.com/mapbox/mapbox-navigation-android/pull/1890) in nav sdk


Chaoba:
> Should we think about removing serializers:

@andrlee  We already removed this method in this PR.

andrlee:
@Chaoba if that's true then that might be a problem since these serializers have to be registered in for nav events to work.

Chaoba:
@andrlee The reason why we need these serializers is because in nav related Events there will be a variable with type `NavigationMetadata` .etc and Gson can't serializers it by default. 

Since we will move all nav events to nav sdk, telemetry should not have any dependency on it and then we need to remove `configureGsonBuilder()` in this PR.

My solution is that we create a NavBaseEvent that will have variable in `NavigationMetadata` in nav sdk and let all other event inherit it, so we will eliminate self define class and Gson could serializer these events by default.

Chaoba:
@andrlee This PR is good to go, will merge to master and test nav sdk with snapshot release.

PR #420, opened 2019-07-25
Add public api to change baseUrl for telemetry

korshaknn:
Problem
Vision team need to change telemetry url at runtime depending on a current country.

Solution
Was decided to add a public API to `mapbox-events-android` lib change telemetry url dynamically.

Test
Unit tests were added to check if a new url is valid and a telemetry client receives setting with a new base url.

closes [1336](https://app.zenhub.com/workspaces/vision-sdk-5ca52e6c027e3e7234ac7512/issues/mapbox/mapbox-vision/1336)

### DIFF
--- a/libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
@@ -18,6 +18,7 @@ import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.regex.Pattern;
 
 import android.support.annotation.NonNull;
 import android.support.v4.content.LocalBroadcastManager;
@@ -475,5 +476,17 @@ public class MapboxTelemetry implements FullQueueCallback, ServiceTaskCallback {
         }
       };
     }
+  }
+
+  @SuppressWarnings("WeakerAccess")
+  public void setBaseUrl(String eventsHost) {
+    if (isValidUrl(eventsHost)) {
+      telemetryClient.setBaseUrl(eventsHost);
+    }
+  }
+
+  private static boolean isValidUrl(String eventsHost) {
+    Pattern urlPattern = Pattern.compile("^[a-z0-9]+([\\-.][a-z0-9]+)*\\.[a-z]{2,5}(:[0-9]{1,5})?(/.*)?$");
+    return eventsHost != null && !eventsHost.isEmpty() && urlPattern.matcher(eventsHost).matches();
   }
 }
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClient.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClient.java
@@ -170,4 +170,9 @@ class TelemetryClient {
 
     return builder.build();
   }
+
+  synchronized void setBaseUrl(String eventsHost) {
+    HttpUrl baseUrl = TelemetryClientSettings.configureUrlHostname(eventsHost);
+    setting = setting.toBuilder().baseUrl(baseUrl).build();
+  }
 }
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClientSettings.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClientSettings.java
@@ -13,15 +13,16 @@ import okhttp3.HttpUrl;
 import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
 
+import static com.mapbox.android.telemetry.MapboxTelemetryConstants.DEFAULT_CHINA_EVENTS_HOST;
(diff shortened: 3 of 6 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClientSettings.java
```diff
@@ -67,6 +70,10 @@ HttpUrl getBaseUrl() {
     return baseUrl;
   }
 
+  void setBaseUrl(String eventsHost) {
```
andrlee: is there a way we could achieve the same with the builder? it feels like this breaks immutability of the `TelemetryClientSettings` class.

libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -476,4 +476,13 @@ public Thread newThread(Runnable runnable) {
       };
     }
   }
+
+  @SuppressWarnings("WeakerAccess")
```
andrlee: i wonder why are we suppressing this lint warning?

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClient.java
```diff
@@ -170,4 +170,8 @@ private RequestBody reverseMultiForm(MultipartBody.Builder builder) {
 
     return builder.build();
   }
+
+  void setBaseUrl(String eventsHost) {
```
andrlee: i suggest to look into how this can be accomplished with the builder.

libtelemetry/src/test/java/com/mapbox/android/telemetry/TelemetryClientTest.java
```diff
@@ -257,6 +257,20 @@ public void checksDebugLoggingEnabledAttachment() throws Exception {
       .debug(eq("TelemetryClient"), contains(" with 1 event(s) (user agent: anyUserAgent) with payload:"));
   }
 
+  @Test
```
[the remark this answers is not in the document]
korshaknn: I can add a regex to check if entered url is valid. Will it be ok?

libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
korshaknn: lint says that access can be package-private. But we need to call it from outside, so make it public and suppress warning

andrlee: now since this is a new public api, i've few questions:
- what happens if this method is called from the main thread?
- what happens if this method is called while the network request is in flight?
- what happens if this method is called from multiple threads at the same time?

libtelemetry/src/test/java/com/mapbox/android/telemetry/TelemetryClientTest.java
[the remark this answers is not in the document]
korshaknn: @andrlee can we remove 
`private static final String HTTPS_SCHEME = "https";`
and add `https` to every default host? 
so in a method setBaseUrl() we will pass a url with `https` prefix and will check if url is valid with regex

because if we pass a string without `https` (and add it later), it is hard to check if the string is valid url

does it make sense?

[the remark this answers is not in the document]
korshaknn: or it is required to set HttpUrl scheme in a separate method?

libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
korshaknn: I think nothing happens if we call it from the main thread.
It won't affect if the request in flight, it will affect the next request.
It's not atomic, so logic could be broken if method is called from multiple threads, but maybe it's an edge case and there is no need to handle it?

or make the method synchronized?
looks like other teams don't need that method, so it will be used only by vision team.

korshaknn:
@andrlee  ptal

libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
andrlee: ok, let's make it at the very east `synchronized`

korshaknn: @andrlee sorry, didn't catch, what does _at the very east_ mean? 
shall I just make it synchronized?

libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -476,4 +477,16 @@ public Thread newThread(Runnable runnable) {
...
+  public void setBaseUrl(String eventsHost) {
+    if (isValidUrl(eventsHost)) {
+      telemetryClient.setBaseUrl(eventsHost);
+    }
+  }
+
+  private boolean isValidUrl(String eventsHost) {
```
andrlee: should this method be `static`?

korshaknn: it is used in a single non-static method, so no need to make it static.

PR #423, opened 2019-08-15
Null pointer fix 

korshaknn:
PR fixes issue [424](https://github.com/mapbox/mapbox-events-android/issues/424)

### DIFF
--- a/libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
@@ -479,8 +479,8 @@ public class MapboxTelemetry implements FullQueueCallback, ServiceTaskCallback {
   }
 
   @SuppressWarnings("WeakerAccess")
-  public void setBaseUrl(String eventsHost) {
-    if (isValidUrl(eventsHost)) {
+  public synchronized void setBaseUrl(String eventsHost) {
+    if (isValidUrl(eventsHost) && checkNetworkAndParameters()) {
       telemetryClient.setBaseUrl(eventsHost);
     }
   }
--- a/libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
@@ -479,10 +479,12 @@ public class MapboxTelemetry implements FullQueueCallback, ServiceTaskCallback {
   }
 
   @SuppressWarnings("WeakerAccess")
-  public synchronized void setBaseUrl(String eventsHost) {
+  public synchronized boolean setBaseUrl(String eventsHost) {
     if (isValidUrl(eventsHost) && checkNetworkAndParameters()) {
       telemetryClient.setBaseUrl(eventsHost);
+      return true;
     }
+    return false;
   }
 
   private static boolean isValidUrl(String eventsHost) {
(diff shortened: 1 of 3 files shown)


libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -479,8 +479,8 @@ public Thread newThread(Runnable runnable) {
   }
 
   @SuppressWarnings("WeakerAccess")
-  public void setBaseUrl(String eventsHost) {
-    if (isValidUrl(eventsHost)) {
+  public synchronized void setBaseUrl(String eventsHost) {
```
yunikkk: I wonder if just ignoring this call is ok, if network is not connected at the moment of the call

korshaknn: network check can be removed, so we can call
`checkRequiredParameters(sAccessToken.get(), userAgent)`

yunikkk: @korshaknn maybe we can return result if base url was set successfully?
I'm afraid with current approach it can be dangerous setBaseUrl was called and ignored and then telemetry will be send to the wrong host

libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -479,10 +479,12 @@ public Thread newThread(Runnable runnable) {
   }
 
   @SuppressWarnings("WeakerAccess")
-  public void setBaseUrl(String eventsHost) {
-    if (isValidUrl(eventsHost)) {
+  public synchronized boolean setBaseUrl(String eventsHost) {
+    if (isValidUrl(eventsHost) && checkNetworkAndParameters()) {
```
nkukday: do you need to check for network here?

korshaknn: `TelemetryClient` is created inside that check. It's a pipeline used by `MapboxTelemetry` class to init `TelemetryClient`

nkukday: `checkNetworkAndParameters()` checks for both network connection and init as required, so if you need the network check then this is fine, if the intent is to check and init `TelemetryClient` then you could  `checkRequiredParameters(..)`

libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -479,10 +479,12 @@ public Thread newThread(Runnable runnable) {
   }
 
   @SuppressWarnings("WeakerAccess")
-  public void setBaseUrl(String eventsHost) {
-    if (isValidUrl(eventsHost)) {
+  public synchronized boolean setBaseUrl(String eventsHost) {
```
nkukday: is `public` safe?

libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
korshaknn: we discussed it internally and decided to check network connection as well.

nkukday:
Approved, could you rebase and merge? Thanks!

Issue #356, opened 2019-03-08
Refactor TelemetryEnabler into passive shared preferences listener


andrlee:
`TelemetryEnabler` accesses shared preferences via `MapboxTelemetry#push` method. While most of the time `push` is accessed from background thread it is not guaranteed that the method will not be misused. 

Refactor `TelemetryEnabler` into a passive shared preferences change listener.

alfwatt:
Do this alongside the dynamic config work

alfwatt:
Valid: https://github.com/mapbox/mobile-telemetry/issues/458

Need decision on PR

PR #426, opened 2019-08-30
Refactor TelemetryEnabler into passive shared preferences listener

harvsu:
This fixes issue #356

### DIFF
--- a/app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
+++ b/app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
@@ -1,6 +1,8 @@
 package com.mapbox.android.events.testapp;
 
 import android.annotation.SuppressLint;
+import android.content.Context;
+import android.content.SharedPreferences;
 import android.location.Location;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
@@ -8,6 +10,8 @@ import android.support.v7.app.AppCompatActivity;
 import android.util.Log;
 import android.view.View;
 import android.widget.Button;
+import android.widget.CompoundButton;
+import android.widget.Switch;
 import android.widget.Toast;
 
 import com.mapbox.android.core.location.LocationEngine;
@@ -20,10 +24,13 @@ import com.mapbox.android.core.permissions.PermissionsManager;
 import com.mapbox.android.telemetry.AppUserTurnstile;
 import com.mapbox.android.telemetry.LocationEvent;
 import com.mapbox.android.telemetry.MapboxTelemetry;
+import com.mapbox.android.telemetry.TelemetryEnabler;
 import com.mapbox.android.telemetry.TelemetryListener;
 
 import java.lang.ref.WeakReference;
 import java.util.List;
+
+import static com.mapbox.android.telemetry.MapboxTelemetryConstants.MAPBOX_SHARED_PREFERENCES;
 
 public class MainActivity extends AppCompatActivity implements PermissionsListener {
   private static final String LOG_TAG = "TelemetryTestApp";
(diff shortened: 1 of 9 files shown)


andrlee:
@harvsu fyi: you could run `build.sh` locally which executes lint, checkstyle, unit and instrumentation tests altogether

alfwatt:
Please squash these changes with `git rebase -i HEAD~3` before merging

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryEnabler.java
```diff
@@ -19,54 +21,98 @@
     ENABLED, DISABLED
   }
 
-  static final String MAPBOX_SHARED_PREFERENCE_KEY_TELEMETRY_STATE = "mapboxTelemetryState";
+
+  // It is intentional to keep this a Class Level variable, so as to keep its life tied to. Otherwise these listeners are stored as WeakReference's and gets garbage collected over time.
```
andrlee: This might be true for activity, not sure if it's true though for any other class that has more deterministic lifecycle.

libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -57,7 +57,7 @@ public MapboxTelemetry(Context context, String accessToken, String userAgent) {
     this.userAgent = userAgent;
     AlarmReceiver alarmReceiver = obtainAlarmReceiver();
     this.schedulerFlusher = new SchedulerFlusherFactory(applicationContext, alarmReceiver).supply();
-    this.telemetryEnabler = new TelemetryEnabler(true);
+    this.telemetryEnabler = new TelemetryEnabler(true, applicationContext);
```
andrlee: i'd rather suggest to deprecate this class going forward. it's extremely fragile and not thread safe.

app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
```diff
@@ -66,6 +73,40 @@ public void onClick(View v) {
...
+    } else {
+
+      telemetrySettingSwitch.setChecked(false);
+    }
+
+    telemetrySettingSwitch.setOnCheckedChangeListener(new CompoundButton.OnCheckedChangeListener() {
+      @Override
+      public void onCheckedChanged(CompoundButton compoundButton, boolean b) {
+
+        if (b) {
```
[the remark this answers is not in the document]
harvsu: Agreed! Will rename this.

app/src/main/java/com/mapbox/android/events/testapp/MainActivity.java
```diff
@@ -66,6 +73,40 @@ public void onClick(View v) {
...
+
+    String telemetryStateName = sharedPreferences
+            .getString(TelemetryEnabler.MAPBOX_SHARED_PREFERENCE_KEY_TELEMETRY_STATE,
+            TelemetryEnabler.State.ENABLED.name());
+
+    if (telemetryStateName.equals(TelemetryEnabler.State.ENABLED.name())) {
+
+      telemetrySettingSwitch.setChecked(true);
+    } else {
+
```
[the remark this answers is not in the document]
harvsu: Agreed! Will reformat .

harvsu:
> I'd suggest to deprecate `TelemetryEnabler` class and implement a new package private version that manages all shared preferences safely. You could steal some ideas from: https://github.com/mapbox/mapbox-events-android/blob/master/libcore/src/main/java/com/mapbox/android/core/crashreporter/MapboxUncaughtExceptionHanlder.java
> 
> On the other note, this change does not address the actual problem, which is in how `MapboxTelemetry` is checking its status by calling `telemetryEnabler.obtainTelemetryState()` which will result in blocking i/o. Lastly, i'm unsure that `static boolean isEventsEnabled(Context context)` is required at all, considering that `KEY_META_DATA_ENABLED` is used only for testing purposes, that said we definitely want to check global status flag before we enable/disable (you can steal some ideas from:
> 
> https://github.com/mapbox/mapbox-events-android/blob/7fd48a4a1bc8cee8dd1163ea1de874af70e9089b/libtelemetry/src/full/java/com/mapbox/android/telemetry/location/LocationCollectionClient.java#L186
> 
> )

@andrlee 
I will look at deprecating TelemetryEnabler per your suggestion.

But, I do not think `telemetryEnabler.obtainTelemetryState()` is a blocking io per current implementation  because it is looking at state from the 'currentTelemetryState' (of TelemetryEnabler instance) variable which is lazily updated listening to the shared preferences.

Per code, isEventsEnabled((Context context) seems to be a higher level setting above TelemetryEnabler that determines if Telemetry can be started/stopped (enable() /disable()). These enable() and disable() calls are referenced in LocationCollectionClient.java which is a non test class.

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryEnabler.java
harvsu: The idea here is to keep the listener's life tied to the TelemetryEnabler instance irrespective of life cycle of the class where the instance is created. For example, if the TelemetryEnabler instance is created in a Non Activity class, the TelemetryEnabler instance will live as long as the parent class/instance is alive.

alfwatt:
Stale

PR #430, opened 2019-10-16
Remove all location permissions from sdk manifest

nkukday:
Problem
For apps targeting <= Android 9 but running on Android 10, the system automatically adds ACCESS_BACKGROUND_LOCATION when updating to Android 10 if either of ACCESS_FINE_LOCATION or ACCESS_COARSE_LOCATION is declared in the manifest.

Solution
Removing all location permissions from SDK to prevent ACCESS_BACKGROUND_LOCATION being added to app's manifest as a result of inheriting location permissions from SDK.

Removing all location permissions from SDK to prevent apps from inheriting location permissions from SDK.

Test
Build [Navigation](https://github.com/mapbox/mapbox-navigation-android) and [Maps](https://github.com/mapbox/mapbox-gl-native-android) successfully.

### DIFF
--- a/libcore/src/main/AndroidManifest.xml
+++ b/libcore/src/main/AndroidManifest.xml
@@ -1,5 +1,4 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="com.mapbox.android.core">
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
 </manifest>
--- a/liblocation/src/main/AndroidManifest.xml
+++ b/liblocation/src/main/AndroidManifest.xml
@@ -1,4 +1,3 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="com.mapbox.android.location">
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
 </manifest>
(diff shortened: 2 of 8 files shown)


langsmith:
cc @mapbox/maps-android

zugaldia:
Instead of removing all location permissions for all API levels, have you considered setting the value of `android:maxSdkVersion` within `uses-permission` so that it only gets included in pre-10 versions?

alfwatt:
@zugaldia see @nkukday's comments: https://github.com/mapbox/mobile-telemetry/issues/445#issuecomment-544612609

Looks like adding the `maxSKDVersion` complicates things for apps which include the SDK, is there anything else blocking approval on this change?

langsmith:
For what it's worth:

I just built https://github.com/mapbox/mapbox-gl-native-android perfectly fine with `4.7.0-SNAPSHOT` for libtelemetry and `1.4.0-SNAPSHOT` for libcore: https://github.com/mapbox/mapbox-gl-native-android/blob/master/gradle/dependencies.gradle#L11-L12

Was able to install the `/mapbox-gl-native-android` test app on my phone and ran a number of the examples. Everything looked fine.

tobrun:
From https://github.com/mapbox/mobile-telemetry/issues/445#issuecomment-544612609:

> The host app will inherit android:maxSdk="28" from us so they will have to explicitly remove this permission and then add their own or they will lose their ACCESS_FINE_LOCATION past API 28.

great catch, this enforces confiugrations on users that they don't expect, :+1: on removal

PR #432, opened 2019-10-30
Prepare annotation processor for navigation implementation

LukasPaczos:
This PR abstracts module generator to be able to:
- provide generic `MapboxModule`s like library loader
- provide specific `NavigationModule`s

We can re-add `MapModule` back when map-specific modules emerge.


annotations-processor/src/main/java/com/mapbox/annotation/processor/Module.kt
```diff
@@ -0,0 +1,12 @@
...
+import javax.lang.model.element.VariableElement
+
+internal data class Module(
+  val name: String,
+  val interfacePackage: String,
+  val interfaceClassName: String,
+  val implPackage: String,
+  val implClassName: String,
+  val constructorParameters: List<VariableElement>
+)
```
tobrun: nit some documentation on this class

LukasPaczos:
After looking at APIs proposed in this PR and implementation from https://github.com/mapbox/mapbox-navigation-android/pull/2243 and @RingerJK's https://github.com/mapbox/mapbox-navigation-android/pull/2240 it becomes visible that limiting constructor access of the custom module implementations could prove to be cumbersome for the developers to use.

As a solution, we can require annotating a **method that returns an implementation instance** of a particular module instead of annotating the whole class. This would push the responsibility for creating the object to the user and offers much more flexibility in an object creation if the implementation has external dependencies. I'm going to address that in this PR.

LukasPaczos:
Reworked the PR to provide an option for module instance injection which increases testability of module implementations.

The generator now outputs 2 types of configuration files. When `skipConfiguration` flag is set to `true` on the annotation:
```
/**
 * Configuration provider for OnboardRouter module.
 */
@Keep
object Mapbox_OnboardRouterModuleConfiguration {
  @JvmStatic
  val skipConfiguration: Boolean = true

  @JvmStatic
  val implPackage: String = "com.mapbox.navigation.route.onboard"

  @JvmStatic
  val implClassName: String = "MapboxOnboardRouter"
}
```
This allows the internal module provider of the SDK to instantiate the module by either calling the no-arg constructor (if available) or injecting params of default Mapbox modules. All default Mapbox modules should have this flag set to `true` and be instantiated internally.

When `skipConfiguration` is set to `false`:
```
/**
 * Configuration provider for OnboardRouter module.
 */
@Keep
object Mapbox_OnboardRouterModuleConfiguration {
  @JvmStatic
  val skipConfiguration: Boolean = false

  /**
   * Set this dependency provider before initializing any components of the modularized library.
   *
   * When you're not using the library anymore, you should pass `null` to clean up the provider
   * reference and prevent memory leaks.
   */
  @JvmStatic
  var moduleProvider: ModuleProvider? = null

  interface ModuleProvider {
    fun createOnboardRouter(): Router
  }
}
```
This requires passing the `ModuleProvider` that returns an instance of the module and enables the user to inject any required module arguments.

PR #435, opened 2019-11-15
Apply android-sdk-versions-plugin to Telemetry and Core modules

harvsu:
This plugin https://github.com/mapbox/android-sdk-versions-plugin validates and persists the SDK version into a file in assets folder. This information further feeds into Reformed User Agent.
https://github.com/mapbox/mobile-telemetry/issues/452
https://github.com/mapbox/mobile-telemetry/issues/423

PR #437, opened 2019-11-21
Add Reformed User Agent to Telemetry Client

harvsu:
Auto generate the user agent to include host application and all Mapbox libraries version information according to the following specification.
mapbox/mobile-telemetry#423
Add New user agent(with header `X-Mapbox-User-Agent`) to network requests sent via TelemetryClient (not ConfigurationClient) in  addition to the existing user agent that is currently in production.

Host Application information is read from package information available at run time and Mapbox library information is available in assets folder (#435). This change reads the information and builds the user agent per format described in the specification.

Note: The new user agent will eventually replace the existing one after the backend is ready to accept this.

### DIFF
--- a/libcore/src/main/java/com/mapbox/android/core/MapboxSdkInfoForUserAgentGenerator.java
+++ b/libcore/src/main/java/com/mapbox/android/core/MapboxSdkInfoForUserAgentGenerator.java
@@ -1,0 +1,88 @@
+package com.mapbox.android.core;
+
+import android.content.res.AssetManager;
+import android.support.annotation.NonNull;
+import android.support.annotation.RestrictTo;
+import android.support.annotation.VisibleForTesting;
+import android.util.Log;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Locale;
+
+/**
+ * Generator that reads(from assets/sdk_versions folder) and constructs Mapbox SDK versions for user agent.
+ * Generates strings in format for each Mapbox library in the host app and concatenates them seperated by spaces.
+ * <p> User agent format for Mapbox SDK : {SDK Name}/{Version} ({packageName}; {versionCode}) </p>
+ */
+public class MapboxSdkInfoForUserAgentGenerator {
+
+  private static MapboxSdkInfoForUserAgentGenerator userAgentGenerator;
+
+  private String sdkInfoForUserAgent;
+  private static final Object lock = new Object();
+  private static final Locale DEFAULT_LOCALE = Locale.US;
+  private static final String USER_AGENT_SDK_VERSION_FORMAT = " %s (%s%s)";
+  private static final String MAPBOX_IDENTIFIER = "mapbox";
+  private static final String EMPTY_STRING = "";
+  private static final String SDK_VERSIONS_FOLDER = "sdk_versions";
+  private static final String LOG_TAG = "MapboxUAGenerator";
+
+  private MapboxSdkInfoForUserAgentGenerator(AssetManager assetManager) {
+    this.sdkInfoForUserAgent = getMapboxSdkIdentifiersForUserAgent(assetManager);
+  }
+
+  public static MapboxSdkInfoForUserAgentGenerator getInstance(@NonNull AssetManager assetManager) {
+    if (userAgentGenerator == null) {
+      synchronized (lock) {
+        userAgentGenerator = new MapboxSdkInfoForUserAgentGenerator(assetManager);
+      }
+    }
+    return userAgentGenerator;
+  }
+
+  @VisibleForTesting
+  @RestrictTo(RestrictTo.Scope.LIBRARY)
+  String getMapboxSdkIdentifiersForUserAgent(@NonNull AssetManager assetManager) {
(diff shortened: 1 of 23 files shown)


libtelemetry/src/test/java/com/mapbox/android/telemetry/TelemetryClientTest.java
```diff
@@ -163,9 +180,10 @@ public void parsesUnauthorizedRequestResponseProperlyPostingAnEvent() throws Exc
 
   @Test
   public void parsesInvalidMessageBodyResponseProperlyPostingAnEvent() throws Exception {
-    MapboxTelemetry.applicationContext = getMockedContext();;
-    TelemetryClient telemetryClient = obtainATelemetryClient("anyAccessToken", "anyUserAgent",
-      MapboxTelemetry.applicationContext);
+    MapboxTelemetry.applicationContext = getMockedContext();
+    ;
```
nkukday: redundant

libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/TestReformedUserAgent.java
```diff
@@ -0,0 +1,24 @@
...
+import org.junit.Test;
+
+public class TestReformedUserAgent {
+
+  @Test
+  public void testReformedUserAgent() {
+    Context context = InstrumentationRegistry.getContext();
+    String reformedUserAgent = TelemetryUtils.createReformedFullUserAgent(context);
+    Assert.assertTrue(reformedUserAgent.contains(context.getPackageName()));
+    Assert.assertTrue(reformedUserAgent.contains("com.mapbox.android.core"));
```
nkukday: possible to extract to local variable, same for other tests

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
```diff
@@ -208,7 +222,7 @@ private static String updateVendorId() {
     return uniqueId;
   }
 
-  private static String obtainApplicationIdentifier(Context context) {
+  static String obtainApplicationIdentifier(Context context) {
```
nkukday: do we want to use this elsewhere?

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
```diff
@@ -221,6 +235,30 @@ private static String obtainApplicationIdentifier(Context context) {
...
+
+  private static String getApplicationName(Context context) {
+    final PackageManager packagemanager = context.getApplicationContext().getPackageManager();
+    ApplicationInfo applicationInfo;
+    try {
+      applicationInfo = packagemanager.getApplicationInfo(context.getPackageName(), 0);
+    } catch (final PackageManager.NameNotFoundException nameNotFoundException) {
+      applicationInfo = null;
+    }
+    return (String) (applicationInfo != null ? packagemanager.getApplicationLabel(applicationInfo) : "(unknown)");
```
nkukday: `(unknown)` can be extracted to a variable

nkukday:
let's update `CHANGELOG.md` with details for this change

libcore/src/main/java/com/mapbox/android/core/MapboxSdkInfoForUserAgentGenerator.java
```diff
@@ -0,0 +1,88 @@
...
+            InputStream inputStream = null;
+            try {
+              inputStream = assetManager.open(SDK_VERSIONS_FOLDER + File.separator + fileName);
+              BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream));
+              String nameAndVersion = reader.readLine();
+              nameAndVersion = nameAndVersion != null ? nameAndVersion : EMPTY_STRING;
+              StringBuilder sdkSubInfo = new StringBuilder(EMPTY_STRING);
+              String subInfo;
+              while ((subInfo = reader.readLine()) != null) {
+                sdkSubInfo.append("; ");
```
alfwatt: Don't include the `; ` for the last line, e.g. `(foo; bar; baz)` is valid but `(foo; bar; baz;)` is invalid

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryClient.java
```diff
@@ -27,21 +27,24 @@
   private static final String EVENTS_ENDPOINT = "/events/v2";
   private static final String ATTACHMENTS_ENDPOINT = "/attachments/v1";
   private static final String USER_AGENT_REQUEST_HEADER = "User-Agent";
+  private static final String USER_AGENT_REQUEST_HEADER_NEW = "X-Mapbox-Agent";
```
alfwatt: Change this to: `MAPBOX_AGENT_REQUEST_HEADER`

libcore/src/main/java/com/mapbox/android/core/MapboxSdkInfoForUserAgentGenerator.java
harvsu: This is not exactly what happens.  ';' is appended before appending any read text from new line only if exists. This is formatted per format "%s%s" where the first %s is the package name and second %s takes the lines that are appended in while loop.

PR #438, opened 2019-11-22
fix: typo displacemnt => displacement

ferdicus:
boyscouting - saw a typo and fixed all occurrences of it


harvsu:
@ferdicus Thanks for the fix and pull request.

PR #445, opened 2019-12-11
Update Node version to >=10

harvsu:
Node 8 gets to EOL in Dec 2019. It is recommended to update to the latest node version as mentioned here:
https://github.com/mapbox/telemetry/issues/572

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/mapbox-events-android",
-  "version": "4.0.0",
+  "version": "4.7.1",
   "description": "Telemetry and Location Services",
   "keywords": [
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/mapbox-events-android",
-  "version": "4.0.0",
+  "version": "4.7.1",
```
nkukday: 4.7.0?

harvsu: 4.7.0 release is already done. Whatever merged to master from now on will be snapshotted at 4.7.1. So, I made this version to be in sync with current telemetry version.

PR #447, opened 2020-01-08
Update firebase test lab config

nkukday:
* Remove redundant devices
* Update artifacts path for `release` since there is no common device between `release` and `build`
* Temporarily ignoring test since issues are inconsistent and we're using different APIs for production and test.

### DIFF
--- a/cloud_test.sh
+++ b/cloud_test.sh
@@ -15,8 +15,7 @@ devices="--device model=m0,version=18,locale=en,orientation=portrait \
 "
 if [ "$is_release" == "true" ]; then
 echo "Is release, adding more devices"
-devices="$devices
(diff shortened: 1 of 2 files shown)


cloud_test.sh
```diff
@@ -13,10 +13,9 @@ devices="--device model=m0,version=18,locale=en,orientation=portrait \
 --device model=sailfish,version=26,locale=en,orientation=portrait \
 --device model=sailfish,version=28,locale=en,orientation=portrait \
 "
-if [ "$is_release" == "true" ]; then
+if [[ "$is_release" == "true" ]]; then
```
harvsu: nit: Redundant additional set of braces.

nkukday: AS Intentions ¯\_(ツ)_/¯

Issue #456, opened 2020-02-20
Obfuscation problem with AppUserTurnstile event


yunikkk:
**Configuration**
 - SDK Version - 4.7.0

**Steps to Reproduce**
We've been facing an [issue](https://app.zenhub.com/workspaces/vision-sdk-5ca52e6c027e3e7234ac7512/issues/mapbox/mapbox-vision/2039) with MAU calculation from Chinese customer (Sogou) for some time. 
Turned out, telemetry library does not expose proguard config to keep [AppUserTurnstile](https://github.com/mapbox/mapbox-events-android/blob/master/libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java) from obfuscation, in fact none of the classes from `com.mapbox.android.telemetry` package are being kept via [proguard config](https://github.com/mapbox/mapbox-events-android/blob/master/libtelemetry/proguard-consumer.pro). 

So if user of some library (eg. Vision), which in turn uses Telemetry is not explicitly keeping telemetry classes from proguard, `AppUserTurnstile` will be obfuscated, and then converted to JSON with wrong schema, eg. 
```
[
  {
    "c": "appUserTurnstile",
    "d": "2020-02-20T18:58:22.326+0300",
    "e": "cbff5455-344f-42ba-96a7-6c542ca3255a",
    "enabled.telemetry": true,
    "g": "B2N_sprout",
    "h": "MapboxVision",
    "i": "0.12.0",
    "j": "Nokia 7 plus",
    "k": "Android - 10",
    "l": null
  }
]
```
Such JSON will then fail to be processed on server so user won't be calculated as MAU. 

BTW, you can notice that `"enabled.telemetry"` is not renamed due to explicit annotation that is used by gson `@SerializedName("enabled.telemetry")`.

Several other mapbox libraries contain specific Telemetry related proguard exclusion configs, eg. https://github.com/mapbox/mapbox-navigation-android/blob/master/libandroid-navigation/proguard-consumer.pro. So users on Navigation won't have issue with MAU calculation.

At the same time I haven't found same config at [Maps proguard](https://github.com/mapbox/mapbox-gl-native-android/blob/master/MapboxGLAndroidSDK/proguard-rules.pro) so I wonder if it's possible that maps users may have similar issues that we have in Vision with missing MAU when obfuscation is turned on? 

Also I've not checked further but it seems other Events may have similar issues with gson serialization.

**Expected**
[...]

PR #457, opened 2020-02-25
Fix Turnstile Obfuscation

harvsu:
This pull request fixes the obfuscation/shrinking in Telemetry SDK that causes malformed Turnstile Events as mentioned in https://github.com/mapbox/mapbox-events-android/issues/456

- Prevent obfuscation of Event classes to keep properties in JSON schema unaltered dfor serialization. Use @Keep annotation(instead of adding in Proguard config as recommended in https://github.com/mapbox/mapbox-gl-native/issues/8978) as a top level setting to keep Event class as is. 

- Throw Assertion Error in Test App in case of POST failure for an event.

- Set minifyEnabled true in build.gradle for debug build to make the test app behave close to production app.

- Add button on test app's main screen, which sends turnstile event on click.

- Add @SerializedName annotation to Turnstile and Crash Event properties to keep them consistent with other events.

- Fix Location Schema test to ignore the properties(speed, course, floor,
speedAccuracy, courseAccuracy, and verticalAccuracy) that are
added in the backend but not incorporated into the SDK yet!


libtelemetry/src/test/java/com/mapbox/android/telemetry/SchemaTest.java
```diff
@@ -56,6 +63,12 @@ public void checkAppUserTurnstileEventFields() throws Exception {
   @Test
   public void checkLocationEventSize() throws Exception {
     JsonObject schema = grabSchema(LOCATION);
+    if (schema != null) {
```
alfwatt: Do we have an issue opened to circle back and modify this test to:
- Read through the `schema` array and ensure that all required fields are present
- Read through the `fields` array and ensure that no fields are missing from the `schema`

harvsu: https://github.com/mapbox/mobile-telemetry/issues/521

libtelemetry/src/main/java/com/mapbox/android/telemetry/AppUserTurnstile.java
```diff
@@ -5,28 +5,39 @@
...
 import static com.mapbox.android.telemetry.TelemetryEnabler.TELEMETRY_STATES;
 
+@Keep
 public class AppUserTurnstile extends Event implements Parcelable {
   private static final String APP_USER_TURNSTILE = "appUserTurnstile";
   private static final String OPERATING_SYSTEM = "Android - " + Build.VERSION.RELEASE;
   private static final String APPLICATION_CONTEXT_CANT_BE_NULL = "Create a MapboxTelemetry instance before calling "
     + "this method.";
 
+  @SerializedName("event")
```
tobrun: `@SerializedName` will be obsolete if you use `@Keep`

harvsu: @SerializedName is added for safety reasons and consistency. 
Still keeping @SerializedName annotation to make sure this does not malfrom the event again in case this class is extended by some other top level class and @Keep annotation is forgotten.

as0577: > `@SerializedName` will be obsolete if you use `@Keep`

PR #443, opened 2019-12-06
Workaround for Strict mode Untagged Socket issue

harvsu:
Untagged socket detection is one of the anomalies detected by Strict Mode in android from Android O. Since the Telemetry Test app enables the Strict mode with penalty death in debug mode, the app crashes/freezes on installation.
Strict mode expects a tag set to every thread and the threads created by Okhttp sockets does not not seem to do so for different reason. This is still an open issue
https://github.com/square/okhttp/issues/3537

This pull request adds custom socket factory and sets TrafficStats.setThreadStatsTag(1000) for configuration client(okhttpClient). The main Telemetry Client has the similar workaround that already in place.
https://github.com/mapbox/mapbox-events-android/blob/22367f5acbbd27183b7c539a5dd8ba96cb59d159/libtelemetry/src/main/java/com/mapbox/android/telemetry/GzipRequestInterceptor.java#L29

PR #464, opened 2020-03-16
Address backwards compatibility issues with AndroidX

harvsu:
**Cherry pick commits in #443 #444 #445 #457**

To address back-porting issues with AndroidX, the current code base is adjusted to build SDKs with and with out AndroidX library.
https://stackoverflow.com/questions/55172932/android-x-backwards-compatibility
To achieve this, the releases are backtracked to Telem 4.7.0/Core 1.4.0 to maintain two different sets of code with and without AndroidX migration changes. 
Telem 4.7.0/Core 1.4.0 is the latest release before which AndroidX was included.

https://github.com/mapbox/mapbox-events-android/tree/long-term-support is branched off at the commit where Telem 4.7.0/Core 1.4.0 release was tagged and this acts as a pseudo master for non AndroidX Code from here on. 

**master** branch tracks the code with AndroidX and includes all new feature/code additions from now on. These new additions will be cherrypicked to https://github.com/mapbox/mapbox-events-android/tree/long-term-support as necessary.

Commits that occurred in releases Telem 4.7.1/Core 1.4.1, Telem 4.7.2 and Teleme 4.7.3(released so far) in master branch will be streamlined and reorganised in further releases as per https://github.com/mapbox/mobile-telemetry/issues/537

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -16,6 +16,9 @@ android {
         release {
             minifyEnabled true
         }
+        debug {
+            minifyEnabled true
+        }
     }
 
     flavorDimensions "version"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,7 +119,7 @@ jobs:
       - deploy:
           name: Publish to Maven Central
           command: |
-            if [[ $CIRCLE_BRANCH == master ]]; then
+            if [[ $CIRCLE_BRANCH == long-term-support ]]; then
               make publish-core ;
               make publish-telem ;
             elif [[ $CIRCLE_TAG == telem-* ]]; then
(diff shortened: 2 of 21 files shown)


app/build.gradle
```diff
@@ -16,6 +16,9 @@ android {
         release {
             minifyEnabled true
         }
+        debug {
```
alfwatt: Won't this make debug builds impossible to debug?

harvsu: Thats right! We had to enable this to address the obfuscation issue. The test app needed to behave close to the production app to detect 422 error because of obfuscated events and we do not build release version for test app.
Locally we have been disabling minify and building for debugging. This has been a bit painful and @nkukday brought this up. 
We would address this by reverting this and adding another debugMinify{minifyEnabled true} to test obfuscated events.

PR #469, opened 2020-03-20
Handle null input while creating LocationEngineResult

nkukday:


liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineResult.java
```diff
@@ -51,8 +53,9 @@ public static LocationEngineResult create(Location location) {
    * @since 1.0.0
    */
   public static LocationEngineResult create(List<Location> locations) {
-    checkNotNull(locations, "locations can't be null");
-    return new LocationEngineResult(locations);
+    return locations != null
```
Guardiola31337: Should we filter out `null` locations from `locations` list? Here we're only checking if the list is `null` 🤔

liblocation/src/test/java/com/mapbox/android/core/location/LocationEngineResultTest.java
```diff
@@ -36,6 +41,49 @@ public void passValidIntent() {
     assertThat(result.getLastLocation()).isSameAs(location);
   }
 
+  @Test
+  public void passNullLocation() {
```
Guardiola31337: Normally is good to structure the tests following the triple A 👀 https://speakerdeck.com/guardiola31337/elegant-unit-testing-mobilization-2016?slide=40 http://wiki.c2.com/?ArrangeActAssert
If the `A`rrange part gets too long we should create factory or builder methods. 

This applies to the rest of the tests 👇

liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineResult.java
```diff
@@ -37,9 +38,10 @@ private LocationEngineResult(List<Location> locations) {
    * @since 1.0.0
    */
   public static LocationEngineResult create(Location location) {
```
Guardiola31337: As this is a `public` method, could we add `@Nullable` / `@NonNull` annotations to help Kotlin users?

liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineResult.java
```diff
@@ -51,8 +53,9 @@ public static LocationEngineResult create(Location location) {
    * @since 1.0.0
    */
   public static LocationEngineResult create(List<Location> locations) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

liblocation/src/test/java/com/mapbox/android/core/location/LocationEngineResultTest.java
```diff
@@ -36,6 +42,85 @@ public void passValidIntent() {
...
+    Location location = null;
```
Guardiola31337: Sounds good, I'd add a new line to explicitly separate `A`rrange, `A`ct and `A`ssert but that's personal preference, I think it reads better. 

👀 [_Working Effectively with Unit Tests_](https://leanpub.com/wewut/read) 
> when writing tests you should prefer DAMP (Descriptive And Maintainable Procedures)

PR #483, opened 2020-06-12
Add default and configurable delay to initialize components

nkukday:
Improving startup time by adding delay for deferred/lazy initialization for some components.

### DIFF
--- a/libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
+++ b/libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
@@ -9,6 +9,7 @@ import android.util.Log;
 import androidx.annotation.NonNull;
 
 import com.mapbox.android.telemetry.errors.ErrorReporterEngine;
+import com.mapbox.android.telemetry.location.LocationCollectionClient;
 
 import java.io.IOException;
 import java.util.Collections;
@@ -32,6 +33,7 @@ import okhttp3.ResponseBody;
 
 import static com.mapbox.android.telemetry.MapboxTelemetryConstants.LOCATION_COLLECTOR_ENABLED;
 import static com.mapbox.android.telemetry.MapboxTelemetryConstants.SESSION_ROTATION_INTERVAL_MILLIS;
+import static com.mapbox.android.telemetry.location.LocationCollectionClient.DEFAULT_SESSION_ROTATION_INTERVAL_HOURS;
 
 public class MapboxTelemetry implements FullQueueCallback, ServiceTaskCallback {
   private static final String LOG_TAG = "MapboxTelemetry";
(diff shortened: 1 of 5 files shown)


libtelemetry/src/full/java/com/mapbox/android/telemetry/MapboxTelemetry.java
```diff
@@ -398,6 +400,9 @@ private synchronized void enableLocationCollector(final boolean enable) {
       @Override
       public void run() {
         try {
+          // lazy install if required
+          LocationCollectionClient.install(applicationContext,
+            TimeUnit.HOURS.toMillis(DEFAULT_SESSION_ROTATION_INTERVAL_HOURS));
```
Chaoba, line 405: Telemetry will be enabled during [map init process](https://github.com/mapbox/mapbox-gl-native-android/blob/238abb470a3774ca924805ee25467382d2e3fb3c/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/module/telemetry/TelemetryImpl.java#L33), this will neutralize the delay initialize. IMHO, it will be better if we put the delay process in `LocationCollectionClient.install`

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
```diff
@@ -173,6 +178,37 @@ public static String generateCreateDateFormatted(Date date) {
...
+    }
+  }
+
+  public static int obtainInitDelay(Context context) {
+    int initDelay = DEFAULT_INIT_DELAY;
+    Bundle metadata = obtainMetaData(context);
+
+    if (metadata != null) {
+      int configuredDelay = metadata.getInt(KEY_META_DATA_INIT_DELAY);
+      if (configuredDelay != 0 && configuredDelay <= MAX_INIT_DELAY) {
```
Chaoba, line 204: Use `configuredDelay > 0` here. If a minus value is passed in, the delay function will not works.

libtelemetry/src/full/java/com/mapbox/android/telemetry/provider/MapboxTelemetryInitProvider.java
```diff
@@ -36,14 +42,30 @@ public boolean onCreate() {
...
+      @Override
+      public void run() {
+        try {
+          LocationCollectionClient.install(getContext(),
+            TimeUnit.HOURS.toMillis(DEFAULT_SESSION_ROTATION_INTERVAL_HOURS));
+        } catch (Exception exception) {
+          Log.e(TAG, exception.getMessage());
+        }
+      }
+    }, initDelay);
```
tobrun, line 66: I'm not sure if I understand the solution here. We will solely delay the initialization, it will still be executed on the main thread?

PR #488, opened 2020-06-18
Add Proguard rules for gms classes

### LINKED ISSUE #487: Reflection used to obtain the Google Fused engine failed when proguard is enabled
**Configuration**
 - SDK Version [telemetry 5.0.1]
 - Android OS Version [10]
 - Device model(s): [Google Pixel 2 XL]
 - App: [Any]

**Steps to Reproduce**
1. Add Google Play Location Services to the classpath
2. Use `LocationEngineProvider` to obtain the `LocationEngine`

**Expected**
`GoogleLocationEngineImpl` is returned

**Actual**
`MapboxFusedLocationEngineImpl` is returned

**Reason**
The `LocationEngineProvider` uses reflection to check whether Google Play Location Services are on the classpath and instantiates the Google's fused location engine if that's the case.
https://github.com/mapbox/mapbox-events-android/blob/9e7b41110a981e020d1d3cebaead1a27568040d1/liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProvider.java#L15-L16
https://github.com/mapbox/mapbox-events-android/blob/9e7b41110a981e020d1d3cebaead1a27568040d1/liblocation/src/main/java/com/mapbox/android/core/location/LocationEngineProvider.java#L47-L48

However, the `liblocation` does not provide consumer proguard rules that would prevent minification of those classes, and the default play services proguard rules do not include them either. This means, that if proguard is enabled, we will always get the `MapboxFusedLocationEngineImpl` back because the class lookup fails, regardless of whether the Google Play Location Services are on the classpath or not.

/cc @mapbox/mobile-telemetry @mapbox/navigation-android

nkukday:
Resolves #487 
Co-authored-by: harvsu <harsha.sura@mapbox.com>

### DIFF
--- a/libcore/proguard-consumer.pro
+++ b/libcore/proguard-consumer.pro
@@ -1,4 +1,6 @@
 # Consumer proguard rules for libcore
 
 # --- GMS ---
+-keep class com.google.android.gms.location.LocationServices
+-keep class com.google.android.gms.common.GoogleApiAvailability
 -dontwarn com.google.android.gms.**


zugaldia:
Thank you for the quick fix!

PR #492, opened 2020-08-10
Fix a crash on newer android OS due to a depreacted API usage

nagineni:
TelephonyManager.getNetworkType() is deprecated in API level 30, so
using getDataNetworkType() to get the current data network type on
newer android OS.

Co-authored-by: Tobrun Van Nuland <tobrun.van.nuland@gmail.com>

### DIFF
--- a/libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/TestCellularNetworkType.java
+++ b/libtelemetry/src/androidTest/java/com/mapbox/android/telemetry/TestCellularNetworkType.java
@@ -1,0 +1,16 @@
+package com.mapbox.android.telemetry;
+
+import android.content.Context;
+import androidx.test.platform.app.InstrumentationRegistry;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestCellularNetworkType {
+  @Test
+  public void testCellularNetworkType() {
+    Context context = InstrumentationRegistry.getInstrumentation().getContext();
+    String cellularNetworkType = TelemetryUtils.obtainCellularNetworkType(context);
+    Assert.assertFalse(cellularNetworkType.isEmpty());
+  }
+}
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
@@ -1,5 +1,6 @@
 package com.mapbox.android.telemetry;
 
+import android.annotation.SuppressLint;
 import android.app.ActivityManager;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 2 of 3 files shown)


alfwatt:
Can we use [hasCarrierPrivledges](https://developer.android.com/reference/android/telephony/TelephonyManager#hasCarrierPrivileges()) to avoid the potential exception in the try/catch block?

alfwatt:
@nkukday @harvsu we're trying to test out this change and appreciate any suggestions or advice, we'll need to get a test app build together and a release to test it out

tobrun:
> Can we use hasCarrierPrivledges to avoid the potential exception in the try/catch block?

That doesn't avoid potential exceptions in the try catch block, see docs on [getNetworkType](https://developer.android.com/reference/android/telephony/TelephonyManager#getDataNetworkType())

```
Requires Permission: READ_PHONE_STATE or that the calling app has carrier privileges (see hasCarrierPrivileges()).
```

The only way to avoid that exception is correctly integration of runtime permissions check. Given that this isn't possible with telemetry setup (no hook in the hosting activity), try-catch is the only thing we can do this.

alfwatt:
@harvsu, @tobrun thanks for the clarifications and suggestion.

@nagineni had questions about how to verify in the test app, @harvsu do we have a step-by-step guild for using the test app to verify a snapshot build of the SDKs?

harvsu:
@alfwatt Unfortunately, we do not have a step by step guild for the test app in this project. But fortunately, it is very straightforward in the main screen. We have leveraged the test app in the past to regression test the location, turnstile events and look for memory leaks. In addition, to test the differences in cross version API's, we used to create necessary build variants in the existing example app for android.

prfarlow1:
So what are the consequences if Android apps using the Mapbox SDK do not have READ_PHONE_STATE permission? Will this library still function correctly?

tobrun:
@prfarlow1 yes, in that case the value of data network will be network type unknown

Abdirahman1234:
So what are the consequences if Android apps using the Mapbox SDK do not have READ_PHONE_STATE permission? Will this library still function correctly?

Abdirahman1234:
So wSo what are the consequences if Android apps using the Mapbox SDK do not have READ_PHONE_STATE permission? Will this library still function correctly?So what are the consequences if Android apps using the Mapbox SDK do not have READ_PHONE_STATE permission? Will this library still function correctly?So what are the consequences if Android apps using the Mapbox SDK do not have READ_PHONE_STATE permission? Will this library still function correctly?hat are the consequences if Android apps using the Mapbox SDK do not have READ_PHONE_STATE permission? Will this library still function correctly?

Abdirahman1234:
Hi I am CABDIRAXMAAN and how it works I need more help. Thanks

PR #498, opened 2020-09-23
[core] Add background location permission

nagineni:
Add background location permission support in the PermissionsManager. Also, request location permissions based on the declarations in the manifest, instead of requesting fine location permission by default.


tobrun:
@nagineni what is left to merge this PR? Would be cool if we could dogfood this downstream

mr1sunshine (approved):

nagineni:
> @nagineni what is left to merge this PR? Would be cool if we could dogfood this downstream

I was waiting for the :heavy_check_mark: We are good to merge this now.

PR #512, opened 2021-02-15
Non-fatal exceptions reporting

friendoye:
This PR:
1) Adds support for reporting non-fatal exceptions via Telemetry SDK (in a similar way to iOS Telemetry SDK);
2) Improve `stackTrace` information: `stackTrace` now keeps more information about original stack trace (more classes might be allowed for keeping in stack trace, exception message also added to stack trace);
3) Fixes failed `SchemaTest`.


tarigo (approved):
LGTM! Thanks, @friendoye!

PR #516, opened 2021-02-25
Improve non-fatal errors reporting

friendoye:
This PR contains small improvements for non-fatal errors reporting:
1. Custom data support was added for non-fatal errors and crashes reporting;
2. Non-fatal errors are not sent immediately (for optimisation purpose).


mr1sunshine (approved):
LGTM

PR #520, opened 2021-05-04
[location] Skip a location update that was determined as worse than the last one.

LukasPaczos:
This skips a location update that was determined as worse than the last one instead of delivering the old location update again, which is an unexpected behavior and causes problems downstream when integrating.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,37 +50,37 @@ jobs:
       - run:
           name: Build all the test APK
           command: ./gradlew assembleAndroidTest
-      - run:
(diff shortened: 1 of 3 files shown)


SiarheiFedartsou (approved):

LukasPaczos:
I still need to adjust tests.

yunikkk (approved):

LukasPaczos:
This should be ready but pending the discussion about this repo's distribution (after Bintray sunset).

tarigo (approved):

PR #546, opened 2021-06-11
[build] Allow artifacts to be overridden for SDK Registry publications.

tarigo:


libtelemetry/build.gradle
```diff
@@ -105,7 +105,7 @@ registry {
     sdkName = project.ext.mapboxRegistrySdkName
     production = true
     snapshot = project.ext.versionName.endsWith("-SNAPSHOT")
-    override = snapshot ? true : false
+    override = true
```
nagineni, line 108: Can we do the same for libcore too?

nagineni (approved):

PR #552, opened 2021-09-06
[build] Attach more info to Maven publications uploaded with sdkRegistry

friendoye:
For `:libcore` and `:libtelemetry` artifacts downloaded from Mapbox Maven repository we cannot retrieve information about license, developers and etc. from POM file. This PR addresses this problem.


tarigo (approved):
Thanks, LGTM!

PR #559, opened 2022-01-20
Fix `TelemetryUtils.obtainApplicationState()` method.

### LINKED ISSUE #557: Android SDK triggers privacy warnings due to Permissions.GET_TASKS
Using com.mapbox.maps:android:10.1.0, I get the following "Security and Trust" warning in Google's Play Console when trying to release my app:

Anomalous Permission Usage
Warning
brightness_1
Privacy
Your app is requesting permissions which are used by less than 1% of functionally similar apps:

android.permission.GET_TASKS 

---

In addition, I had to disclose the collection of installed apps in the Play Console's data safety section.  I received an email from Google telling me that my data safety report was incorrect:

We found an issue in the following area(s):

Policy Declaration for Play Safety Label: App Activity Data Type - Installed Apps
Starting in February 2022, Data safety sections reviewed and approved will start appearing to users; make any changes needed by then. Starting April 2022, your app must be in compliance with this policy or app updates will be rejected. If your app continues to be non-compliant beyond April 2022 your app may face additional enforcement actions in the future.

---

So every app that uses Mapbox will have to declare the collection (and possibly reporting?) of installed apps, or they won't be able to publish updates after April.

I found that this permission is coming from the following Mapbox library:
com.mapbox.mapboxsdk:mapbox-android-telemetry:8.1.0

I'm not sure why it would need to know what other apps are installed.  This seems invasive at best, and much more serious if this is actually being sent off the phone, which apps have no way of knowing. 

Can this permission be removed from the telemetry library?

tarigo:
Fixes #557.

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
@@ -1,7 +1,6 @@
 package com.mapbox.android.telemetry;
 
 import android.annotation.SuppressLint;
-import android.app.ActivityManager;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
@@ -25,7 +24,6 @@ import java.net.UnknownHostException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


kiryldz (commented):
We need to remove GET_TASKS permission from manifest file as well

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
```diff
@@ -121,20 +119,18 @@ public static String obtainUniversalUniqueIdentifier() {
   }
 
   public static String obtainApplicationState(Context context) {
-    ActivityManager activityManager = (ActivityManager) context.getSystemService(Context.ACTIVITY_SERVICE);
```
pengdev, line 124: Shall we also remove the permission from the AndroidManifest.xml?
```
    <uses-permission android:name="android.permission.GET_TASKS" android:maxSdkVersion="20"/>
```

tarigo: Sure!

tarigo:
> We need to remove GET_TASKS permission from manifest file as well

Yep, done.

libtelemetry/src/main/java/com/mapbox/android/telemetry/TelemetryUtils.java
```diff
@@ -121,20 +119,18 @@ public static String obtainUniversalUniqueIdentifier() {
...
-    List<ActivityManager.RunningAppProcessInfo> appProcesses = activityManager.getRunningAppProcesses();
-    if (appProcesses == null) {
-      return NO_STATE;
+    String state = NO_STATE;
+    switch (AppStateUtils.getAppStateFromActivityManager(context)) {
```
kiryldz, line 123: I may not see full picture but couldn't we actually expose and use `AppStateUtils.getAppStateFromActivityManager(context)` instead of `obtainApplicationState`? This mapping here seems to be actually almost 1 to 1 (at least looking at the diff here).

tarigo: Historical reasons. :)  We can expose `AppStateUtils.getAppStateFromActivityManager(context)` but I would avoid removing `obtainApplicationState` cause it's a breaking change.

kiryldz: Couldn't we do something like
```
public static String obtainApplicationState(Context context) {
  return appStateUtils.getAppStateFromActivityManager(context);
}
```
if I get the mapping correctly?

tarigo: Well, NO_STATE is an empty string, but others are one-to-one, simplified.

kiryldz (approved):
Thanks @tarigo, looks good to me now 👍

PR #566, opened 2022-05-31
[location] Add new field `permissionStatus` to location events

mr1sunshine:
With this change, we are adding extra information about granted location permission to the location events:

* `AllowAlways` means that the end-user is allowed to collect location information while an application is in the background.
* `AllowWhenInUse` means that the end-user is allowed to collect location information while an application is in the foreground or only one-time access was granted.

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationEvent.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationEvent.java
@@ -35,8 +35,11 @@ public class LocationEvent extends Event implements Parcelable {
   private String applicationState;
   @SerializedName("horizontalAccuracy")
   private Float accuracy = null;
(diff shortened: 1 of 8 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationEvent.java
```diff
@@ -35,8 +35,11 @@ public class LocationEvent extends Event implements Parcelable {
   private String applicationState;
   @SerializedName("horizontalAccuracy")
   private Float accuracy = null;
+  @SerializedName("permissionStatus")
+  private final String permissionStatus;
 
-  public LocationEvent(String sessionId, double latitude, double longitude, String applicationState) {
```
pengdev, line 39: nit: as this is a public API, adding a new parameter would be a breaking change.

PR #567, opened 2022-06-01
[libtelemetry] Fix LocationEvent timestamp.

tarigo:
Take the timestamp of the location fix not from a system clock, but from the fix itself (i.e. value populated by system geolocation service).


JamesSaxon:
Thanks @tarigo, this is great to see!  Any sense of how long it will take for this to reach devices and filter through to the data?

tarigo:
> Thanks @tarigo, this is great to see! Any sense of how long it will take for this to reach devices and filter through to the data?

Probably few weeks. It was release today, so as soon as the new version is adopted we should start getting correct data (I hope :sweat_smile:).

PR #569, opened 2022-06-07
[libtelemetry] Add `LocationEvent` constructor that doesn't accept `permissionStatus`

mr1sunshine:
In #566 we added `permissionStatus` parameter to the `LocationEvent` constructor. That change actually introduced a breaking change in the public API surface.
With this change, we are bringing an old constructor back and we use the `unknown` default value for the `permissionStatus`.

### DIFF
--- a/libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationEvent.java
+++ b/libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationEvent.java
@@ -57,6 +57,12 @@ public class LocationEvent extends Event implements Parcelable {
     this(sessionId, latitude, longitude, System.currentTimeMillis(), applicationState, permissionStatus);
   }
 
+  @Deprecated
+  public LocationEvent(String sessionId, double latitude, double longitude, long timestampUtcMs,
+                       String applicationState) {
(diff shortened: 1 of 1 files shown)


libtelemetry/src/main/java/com/mapbox/android/telemetry/LocationEvent.java
```diff
@@ -38,6 +38,11 @@ public class LocationEvent extends Event implements Parcelable {
   @SerializedName("permissionStatus")
   private final String permissionStatus;
 
+  public LocationEvent(String sessionId, double latitude, double longitude, long timestampUtcMs,
```
pengdev, line 41: should we also mark this as deprecated and use the the one with permissionStatus by default?

mr1sunshine: @pengdev I was thinking about that. Will it cause compilation warnings/errors for customers?

pengdev: no, it will be a lint warning.

mr1sunshine: done

tarigo (approved):

pengdev (approved):

PR #571, opened 2022-06-30
[telemetry] Remove WAKE_LOCK permission

zach2good:
We no longer want to have to specify the WAKE_LOCK permission, so we need to replace JobIntentService with Worker and WorkManager, which doesn't have this dependency.


DorkMatter (approved):

Guardiola31337:
@zach2good 
> Had to use WorkManager "2.6.0", since "2.7.1" was forcing a Java 1.8 dependency.

Noting that for apps that don't depend on WorkManager OP changes may cause:

```
2022-07-08 09:31:51.741 23122-23155/com.mapbox.navigation.examples E/AndroidRuntime: FATAL EXCEPTION: pool-7-thread-1
    Process: com.mapbox.navigation.examples, PID: 23122
    java.lang.IllegalArgumentException: com.mapbox.navigation.examples: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
    Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.
        at android.app.PendingIntent.checkFlags(PendingIntent.java:375)
        at android.app.PendingIntent.getBroadcastAsUser(PendingIntent.java:645)
        at android.app.PendingIntent.getBroadcast(PendingIntent.java:632)
        at androidx.work.impl.utils.ForceStopRunnable.getPendingIntent(ForceStopRunnable.java:285)
        at androidx.work.impl.utils.ForceStopRunnable.isForceStopped(ForceStopRunnable.java:158)
        at androidx.work.impl.utils.ForceStopRunnable.forceStopRunnable(ForceStopRunnable.java:185)
        at androidx.work.impl.utils.ForceStopRunnable.run(ForceStopRunnable.java:103)
        at androidx.work.impl.utils.SerialExecutor$Task.run(SerialExecutor.java:91)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:920)
2022-07-08 09:31:51.742 23122-23145/com.mapbox.navigation.examples E/FirebaseCrashlytics: Tried to write a fatal exception while no session was open.
```

Run into ^^ when bumping to `8.1.4` from the Navigation SDK https://github.com/mapbox/mapbox-navigation-android/pull/6026

In the Navigation SDK we had to explicitly include that dependency https://github.com/mapbox/mapbox-navigation-android/blob/5b32ce720229657d1da67d4aa42732d3fa116f47/libnavigation-metrics/build.gradle#L30-L31 to avoid above-mentioned runtime crash.

BTW Java 1.8 shouldn't be a problem as many libraries require these options in `build.gradle` nowadays:

```groovy
compileOptions {
    sourceCompatibility = JavaVersion.VERSION_1_8
    targetCompatibility = JavaVersion.VERSION_1_8
}
```

cc @Zayankovsky

Guardiola31337:
Wondering why Maps didn't run into this https://github.com/mapbox/mapbox-maps-android/pull/1482 🤔 

cc @ank27 @yunikkk

yunikkk:
@Guardiola31337 will recheck in on maps side

Guardiola31337:
PR https://github.com/mapbox/mapbox-events-android/pull/573 addressing https://github.com/mapbox/mapbox-events-android/pull/571#issuecomment-1179010463

yunikkk:
> Wondering why Maps didn't run into this [mapbox/mapbox-maps-android#1482](https://github.com/mapbox/mapbox-maps-android/pull/1482) 🤔
> 
> cc @ank27 @yunikkk

Passed unnoticed in Maps since reproduces only when targetSdk = 31+ and device is also 31+, and we've used target 30 everywhere. Will enhance testing pipelines with different targetSdk's to catch such issues in future. 
Btw do you do similar kind of tests (different targetSdk lvl) in Nav?

Guardiola31337:
@yunikkk noting that https://github.com/mapbox/mapbox-events-android/releases/tag/telem-8.1.5-core-5.0.2 went out so you should be good after bumping into `8.1.5`.

> Btw do you do similar kind of tests (different targetSdk lvl) in Nav

On Nav, we try to be up-to-date when it comes to the `targetSdkVersion` / `compileSdkVersion` as new versions are available. For instance, we bumped to `targetSdkVersion` `31` some time ago https://github.com/mapbox/mapbox-navigation-android/pull/5259 😅

yunikkk:
> For instance, we bumped to `targetSdkVersion` `31` some time ago [mapbox/mapbox-navigation-android#5259](https://github.com/mapbox/mapbox-navigation-android/pull/5259) 😅

Suppose could still expect issues with target 32, and soon-ish with 33 =)

Guardiola31337:
@yunikkk could you share the PR? We should have similar tests on the Navigation SDK.

alexander-kulikovskii:
> @yunikkk could you share the PR? We should have similar tests on the Navigation SDK.

https://github.com/mapbox/mapbox-maps-android-internal/pull/1238

yunikkk:
> @yunikkk could you share the PR? We should have similar tests on the Navigation SDK.

🤒 I was going to share it initially, thanks @alexander-kulikovskii for the link

